PR #15, opened 2022-02-04
Container Controller Makefile

kon14:
Before testing out `get-conduit.sh`, make sure you edit the file and set `TAG=docker` as `master` won't work before merge.

Currently hardcodes v0.10.7 for Conduit-UI.
Works around <v0.11 config restart issue using sleep, but ideally we'd be health checking the db instead.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,26 +1,443 @@
-##
-# Conduit Docker Compose
-#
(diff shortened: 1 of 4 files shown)


kkopanidis (approved):
@kon14 works like a charm

PR #21, opened 2022-02-08
feat(cms): remove cms-only filtering from getSchema, getSchemas

kon14:
CMS.getSchema and CMS.getSchemas now also return schemas not-owned by CMS.
CustomEndpoints remains blacklisted and therefore won't show up.

This PR also drops getSchemasFromOtherModules route as it's now pointless and unused in production, therefore branding as Breaking Change and deprecating straight away to avoid cluttering up the codebase.
@KostasFeg I believe Conduit-UI makes use of it to some extent?

BREAKING CHANGE:
Removed CMS.getSchemasFromOtherModules (GET: /admin/cms/schemasFromOtherModules)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

If yes, please describe the impact and migration path for existing applications:

- [x] It's submitted to the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx`, where "xxx" is the issue number)

**Other information:**

### DIFF
--- a/modules/cms/src/admin/schema.admin.ts
+++ b/modules/cms/src/admin/schema.admin.ts
@@ -31,7 +31,10 @@ export class SchemaAdmin {
   }
 
   async getSchema(call: ParsedRouterRequest): Promise<UnparsedRouterResponse> {
-    const query: {[p: string]: any} = { _id: call.request.params.id, name: { $nin: CMS_SYSTEM_SCHEMAS } };
+    const query: {[p: string]: any} = {
+      _id: call.request.params.id,
+      name: { $nin: CMS_SYSTEM_SCHEMAS }
+    };
     if (!isNil(call.request.params.owner)) {
       query.ownerModule = call.request.params.owner;
     }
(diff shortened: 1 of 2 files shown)


modules/cms/src/admin/schema.admin.ts
```diff
@@ -71,31 +80,17 @@ export class SchemaAdmin {
...
-      ownerModule: 'cms', name: { $nin: CMS_SYSTEM_SCHEMAS }
-    });
-    const schemaNamesFromCMS = (schemasFromCMS as _DeclaredSchema[]).map(
-      (schema: _DeclaredSchema) => schema.name
-    );
-    const schemasFromOtherModules = allSchemas.filter((schema: any) => {
-      return !schemaNamesFromCMS.includes(schema.name);
-    });
+    for (const schema of schemas) {
+      if (schema.ownerModule !== 'cms') { // || CMS_SYSTEM_SCHEMAS.includes(schema.name)) {
```
kon14, line 84: Leaving this comment in case we wish to re-enable CustomEndpoints later on.

modules/cms/src/admin/schema.admin.ts
```diff
@@ -71,31 +80,17 @@ export class SchemaAdmin {
...
+        schema.modelOptions.conduit.cms = {
+          enabled: true,
+          crudOperations: false,
+          authentication: false, // here due to type restriction
```
kon14, line 88: This doesn't look particularly clean imo. Maybe Conduit-UI should contain this logic instead?

kkopanidis: why would this be added to every schema? These values make no sense when the schema isn't owned by the CMS. They won't be able to alter them so why include them?

kon14: Technically it shouldn't. Conduit-UI assumes it does, but yeah, I agree it shouldn't be here.
I'll remove this and ping for review/merge after confirming Conduit-UI works as expected.

PR #41, opened 2022-02-17
Authenticate with phone #12w6xv4

sdimitris:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The feature that this PR brings, is the phone authentication. 
A user by using only its phone number can register or login.


modules/authentication/src/routes/routes.ts
```diff
@@ -176,16 +192,37 @@ export class AuthenticationRoutes {
...
+      }
       if (authConfig?.twofa.enabled) {
         this._routingController.route(
           {
             path: '/local/twofa',
             action: ConduitRouteActions.POST,
             description: `Verifies the 2FA token.`,
             bodyParams: {
-              email: ConduitString.Required,
+              email: ConduitString.Optional,
```
kkopanidis, line 225: email is required

modules/authentication/src/routes/routes.ts
```diff
@@ -106,6 +106,22 @@ export class AuthenticationRoutes {
         this.localHandlers.authenticate.bind(this.localHandlers),
       );
 
+      this._routingController.route(
```
kkopanidis, line 109: why isn't this route inside the IF block that checks for phone authentication?

modules/authentication/src/handlers/local.ts
```diff
@@ -643,6 +741,14 @@ export class LocalHandlers {
       console.log('sms 2fa not active');
     }
 
+    if ((config.phoneAuthenticate.enabled) && !errorMessage) {
```
kkopanidis, line 744: phone authentication is not part of the "local" strategy

modules/authentication/src/routes/routes.ts
```diff
@@ -106,6 +106,22 @@ export class AuthenticationRoutes {
...
+        {
+          path: '/local/phone',
```
kkopanidis, line 111: phone authentication is not part of the /local strategy

modules/authentication/src/handlers/local.ts
```diff
@@ -216,6 +217,86 @@ export class LocalHandlers {
...
+    const config = ConfigController.getInstance().config;
+    if (isNil(context))
+      throw new GrpcError(status.UNAUTHENTICATED, 'No headers provided');
+    let user: User | null = await User.getInstance().findOne(
+      { phoneNumber: phone },
+    );
+    if (isNil(user)) {
+      user = await User.getInstance().create({
+        phoneNumber: phone,
+        isVerified: false,
```
kkopanidis, line 235: verified should be true. As with the email login should not be permitted for a user if their credentials are not verified.

kkopanidis (changes requested):

modules/authentication/src/routes/routes.ts
sdimitris: Changed it by mistake.

modules/authentication/src/handlers/phone.ts
```diff
@@ -0,0 +1,174 @@
...
+        },
+      },
+      new ConduitRouteReturnDefinition('LoginResponse', {
+        message: ConduitString.Required,
+      }),
+      this.authenticate.bind(this),
+    );
+    this.routingManager.route(
+      {
+        path: '/phone-login',
```
kkopanidis, line 64: make it something like /phone/verify, since both endpoints work both for login and register. So maybe the first one should be /phone and the second one /phone/verify or /phone/authenticate

kkopanidis (changes requested):

PR #62, opened 2022-03-03
Routing Improvements

kkopanidis:
The target of this PR is to make routing declaration more consistent and clear older code and methods of declaring routes.
Goals:
- Consistent declaration and features between Admin and Client routers
- Remove custom code from Admin/Client router grpc client files
- Create any and all utility functions required to add enough sugar in route declaration

- [ ] Bugfix
- [ x Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

If yes, please describe the impact and migration path for existing applications:
Existing applications should use the new functions introduced and remove the older functions. Applications using the 0.11 sdk version should not be affected.

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx`, where "xxx" is the issue number)

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**


kkopanidis:
I'm still thinking that maybe I need to also make the admin changes in this PR, so that admin and client router registrations are consistent but I'll leave it to the reviewers to decide if the PR is already too big

modules/forms/src/controllers/forms.controller.ts
```diff
@@ -36,10 +35,10 @@ export class FormsController {
       });
   }
 
-  refreshRoutes() {
+  private async loadExistingForms() {
```
kon14, line 38: Why do we need a private copy of refreshRoutes?

kon14 (changes requested):
Can we keep a consistent naming for RoutingManager object members?
Some classes use _routingController while others use _routingManager.

Besides that, I think we can get rid of `loadExistingForms()` in `Forms.FormsController` and call the public `refreshRoutes` in its constructor just like we do everywhere else.

Beyond that, looks LGTM 🎉

kon14:
> I'm still thinking that maybe I need to also make the admin changes in this PR, so that admin and client router registrations are consistent but I'll leave it to the reviewers to decide if the PR is already too big

Lets address that in a followup PR. Prettier already produces huge diffs.

kon14 (approved):

sdimitris (approved):

Michael-Vol (approved):

PR #82, opened 2022-03-19
fix(core): socket pushes not working

kkopanidis:
The io object seems weird in its behaviour,
but the changes should allow for connected
sockets to receive the data correctly

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:


kkopanidis:
Proper testing is needed, I'm still facing network issues.

modules/chat/src/routes/routes.ts
```diff
@@ -299,7 +299,7 @@ export class ChatRoutes {
 
   async onMessage(call: ParsedSocketRequest): Promise<UnparsedSocketResponse> {
     const { user } = call.request.context;
-    const [roomId, message] = call.request.params;
+    const { roomId, message } = call.request.params;
```
kon14, line 302: This should be `[ roomId, message ]`. Socket params are an array.

modules/chat/src/routes/routes.ts
```diff
@@ -323,7 +323,7 @@ export class ChatRoutes {
 
   async onMessagesRead(call: ParsedSocketRequest): Promise<UnparsedSocketResponse> {
     const { user } = call.request.context;
-    const [roomId] = call.request.params;
+    const { roomId } = call.request.params;
```
kon14, line 326: Same ^

kon14 (changes requested):
ParsedSocketRequest's `call.request.params` are an array.
Object destructuring will produce `undefined`.

kkopanidis:
@kon14 slow down, I'm looking at these right now, I only resolved conflicts previously. Even without the resolution it wouldn't work there are multiple typing issues.

kon14 (approved):

PR #83, opened 2022-03-20
fix(router, admin): Admin Swagger headers and route paths

### LINKED ISSUE #80: [BUG] Admin Swagger uses User Security Headers
**Describe the bug**
SwaggerGenerator currently hardcodes the following non-admin headers as its header security scheme:
- security client id
- security client secret
- Bearer user token

Admin routes require masterkey and a JWT prefixed admin token instead.

**To Reproduce**
1. Go to `/admin/swagger`
2. Click on 'Authorize'
3. Notice the incorrect headers

**Expected behavior**
Each Swagger instance should expose the appropriate headers.
Pass a securityScheme object to SwaggerGenerator through RestController.

kon14:
Closes #80 and generally improves Swagger route doc generation.
- Admin routes are now properly prefixed by `/admin/` so that Swagger users no longer need to guess the proper baseUrl.
- Admin routes now require proper admin headers.

This PR also converts Admin router's `Authentication` header format from `JWT adminToken` to `Bearer adminToken`.
This change improves compatibility with OpenAPI and Swagger, while also improving consistency with user authentication.

This bit is not a breaking change as backwards compatibility has been preserved.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx`, where "xxx" is the issue number)

### DIFF
--- a/packages/router/src/controllers/Rest/Swagger.ts
+++ b/packages/router/src/controllers/Rest/Swagger.ts
@@ -3,9 +3,11 @@ import { SwaggerParser } from './SwaggerParser';
 import { isNil } from 'lodash';
 
 export class SwaggerGenerator {
+  private readonly _swaggerDoc: any;
+  private readonly _globalSecurityHeaders: any;
   private _parser: SwaggerParser;
 
-  constructor() {
+  constructor(securitySchemes: any, globalSecurityHeaders: any) {
     this._swaggerDoc = {
       openapi: '3.0.0',
       info: {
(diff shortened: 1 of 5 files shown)


kon14:
`any` types due to pre-existing routeDoc missing a type. Could technically add a type for that one tho.

packages/router/src/controllers/Rest/Swagger.ts
```diff
@@ -69,12 +67,7 @@ export class SwaggerGenerator {
           },
         },
       },
-      security: [
-        {
-          clientid: [],
-          clientSecret: [],
-        },
-      ],
+      security: JSON.parse(JSON.stringify(this._routerMetadata.globalSecurityHeaders)),
```
kon14, line 70: We could also entirely avoid repeated stringifications by just passing a pre-stringified `globalSecurityHeaders` instead. Totally type hacky, but definitely worth it perf-wise.

kon14: Ended up stringifying once per Swagger construction instead (that makes 2 times total).

kkopanidis (approved):

PR #103, opened 2022-03-30
feat(authentication,config): remove Authentication's dependency on Email for unverified local email auth strategies

### LINKED ISSUE #99: [FEAT] Authentication module's initial run should not depend on Email or further configuration

Most deployments require some form of authentication.
In our case, that pretty much translates to having the Authentication module available.
Therefore, most users wishing to try out Conduit for the first time are most likely not going to be head over heels with the concept of spending much time on configuring it just so they can move on to finally giving the project a spin.

The module's initial configuration currently enables the "local" authentication strategy.
This is fairly convenient and greatly facilitates initial setup as it involves no third party providers.

The issue arises from "local" authentication defaulting to an email identifier, which results in Authentication full initialization effectively depending on either:

1. the user updating Authentication's config to make use of a username-based identifier
2. the Email module being available **and properly configured**

Option 1) requires a one-time administrative configuration request. Not too bad, until you consider how this affects our documentation and overcomplicates what would otherwise be fairly simple first timer tutorials.

Regarding 2), the problem is only magnified, as choosing to actually go with an email identifier not only requires that the Email module is brought online and configured, but also results in users having to possibly create an account at an email provider and specify all their creds in the module's configuration.
Unavoidable for actual deployments, but definitely overkill for trying out the project.

The Authentication module should not depend on the Email module being configured nor available on initial run.
I can think of a couple different ways to achieve this.
Regardless of which one we end up going with, we should keep defaulting to the "local" authentication.

a) Have "local" auth strategy default to a "username" identifier. Extremely easy to do, but username auth is non ideal.

b) Keep defaulting to "email", switching off "sendVerificationEmail" and "verificationRequired", then only have Authentication actually wait on Email if said options are actually enabled so that the barebones "email" identifier does not actually depend on the Email module.

c) Keep the default config file as is, adding an env var to override the default identifier on a non-configured instance. Kinda needlessly complicated, plus the user is also going to be aware of the env var. What if they just launch Auth once without it and have the database store the "email" identifier config etc.

I'm leaning towards b).

kon14:
This PR greatly improves the experience of setting up a new deployment from scratch by removing Authentication`s Email dependency on a fresh database. Fixes #99.

This change is going to be extremely useful for newcomers trying out Conduit for the first time as the docs can now be simplified in that regard.
As far as backend contributors are concerned, it's also going to significantly reduce the amount of time spent on submitting configuration requests just to get an active local auth post db wipes.

The Authentication's configuration schema's local section has also been slightly altered so as to keep things clean and structured.
Check out the relevant changes below:

Old:
```
local: {
  identifier: string,
  enabled: boolean,
  sendVerificationEmail: boolean,      // changed
  verificationRequired: boolean,       // changed
  verification_redirect_uri: string,   // changed
  forgot_password_redirect_uri: string,
},
```

New:
```
local: {
  identifier: string,
  enabled: boolean,
  verification: {
    required:boolean,
    sendEmail: boolean,
    redirectUri: string,
  },
  forgot_password_redirect_uri: string,
},
```

As far as I can tell, the `sendEmail` part is also redundant at this point, as we don't currently support different forms of verification.
We could most likely just get rid of it too.

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes
- [ ] No

Due to the Config module being the one to perform the migration (ownership constraints),
**--> Core has to be restarted before Authentication <--**
so that the latter`s config file can be properly updated.

- [X] When resolving a specific issue, it's referenced in the PR's title

- [X] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

### DIFF
--- a/modules/authentication/src/Authentication.ts
+++ b/modules/authentication/src/Authentication.ts
@@ -122,7 +122,16 @@ export default class Authentication extends ManagedModule {
   async userCreate(call: any, callback: any) {
     const email = call.request.email;
     let password = call.request.password;
-    let verify = call.request.verify;
+    const verify = call.request.verify;
+
+    const verificationConfig = ConfigController.getInstance().config.local.verification;
+    if (verify && !(verificationConfig.required && verificationConfig.sendEmail)) {
+      return callback({
+        code: status.INVALID_ARGUMENT,
+        message: 'Email verification is disabled. Configuration required.',
+      });
+    }
+
     if (isNil(password) || password.length === 0) {
       password = AuthUtils.randomToken(8);
     }
--- a/modules/authentication/src/config/config.ts
+++ b/modules/authentication/src/config/config.ts
@@ -15,20 +15,21 @@ export default {
       format: 'Boolean',
       default: true,
     },
-    sendVerificationEmail: {
-      doc:
-        'Defines if the authenticator should automatically send a verification e-mail to the user',
-      format: 'Boolean',
-      default: true,
-    },
-    verificationRequired: {
-      doc: 'Defines if email verification is required for login',
-      format: 'Boolean',
-      default: true,
-    },
-    verification_redirect_uri: {
-      format: 'String',
-      default: '',
+    verification: {
+      required: {
+        doc: 'Defines if email verification is required for login',
+        format: 'Boolean',
+        default: false,
+      },
+      sendEmail: {
+        doc: 'Defines if the authenticator should automatically send a verification e-mail to the user',
+        format: 'Boolean',
+        default: true,
+      },
+      redirectUri: {
+        format: 'String',
+        default: '',
+      },
     },
     forgot_password_redirect_uri: {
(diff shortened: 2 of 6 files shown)


kon14:
On second thought, going to improve on this prior to merge in the following ways:

- Move the migration to Authentication so as to remove the need for guided update
- Phase out redundant "username" local auth identifier (breaking change, no handler-level compat)
- Add a configurable option for allowing/preventing user self-registration
- Migrate config fields to camel_case (we should standardize this across our configs)

kon14:
This PR also introduces proper character validation for user email fields.

sdimitris (approved):

Michael-Vol (approved):

kkopanidis (approved):

PR #113, opened 2022-04-06
Update GitHub Issue Templates

kon14:
This updates our issue templates and converts them from Markdown to the new YAML issue forms format.
Required fields are now actually required, with proper placeholders, sections are clearly distinguished, no longer indicates stuff through markdown comments etc.

It also customizes our template chooser with links to Conduit UI, the Docs and our Discord `help` channel.
I disabled blank issues as I think the existing options pretty much cover mostly everything.
Do you think I should revert that?

Thoughts? Should we still keep a separate `Question` template for anyone that's not into Discord?

You can check it out live [here](https://github.com/kon14/ConduitTestingChamber/issues/new/choose).

Here's a preview of how it looks like:

Screenshot from 2022-04-07 00-03-38

Screenshot from 2022-04-07 00-05-44


kon14:
We could also probably get rid of the `[Bug]:`/`[Feature]:` issue name prefixes and stick to tags.
That would prevent confusion regarding any issues moved to a different category.

Michael-Vol (approved):

kkopanidis (approved):

Issue #116, opened 2022-04-09
[Feature]: Users should not be capable of unilaterally adding other users to any form of group


kon14:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/ConduitPlatform/Conduit/blob/main/.github/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [Issue Tracker](https://github.com/ConduitPlatform/Conduit/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Opening this issue so as to document the way we currently approach user group related "add user" functionality and provide some suggestions on how we can improve on it for both existing and upcoming module features.

The limitation I'm about to describe is currently exclusively related to the Chat module's chat group feature, but should soon start affecting Authentication once its own user groups feature gets released.

As of right now, user authenticated application requests for creating a chat room and adding additional users to it unilaterally add target users to the group without any way for them to accept or decline this action.

While administrators should definitely be capable of operating in such a way, normal user requests should ideally be inviting other users to a group instead.
Automatically adding others into a group can not only end up being frustrating, depending on the app, but may also end up getting exploited by malicious users spamming such requests.
We would be retaining the existing functionality for apps that are best suited for it through a configuration option of course.

Adding users to a group, in the default configuration, should ideally send them an invitation for them to accept or decline.
We should also be sending out socket events for this so that client apps can pick these up in real time.

The very same considerations should be taken into account while implementing Authentication's user groups feature.

If we wish to take this up a notch at some point down the line we could make it so users' connections of sorts can be allowed to automatically add them to groups or even let users themselves configure their own preferences, but offering simple module configurations for it would be more than enough for most use cases and definitely not overkill.

PR #119, opened 2022-04-15
Chat room invitation

sdimitris:
Solves the issue #116 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

The PR introduces invitation strategy on the chat rooms.
A client can either accept or decline an invitation if the module is configured to do this.

- Chat Configuration changed to ->
```   
sendInvitations: {
    enabled: {
      doc: 'Defines if this strategy is active or not',
      format: 'Boolean',
      default: false,
    },
    send_email: {
      doc: 'Defines if the sender should automatically send an invitation e-mail to the user',
      format: 'Boolean',
      default: false,
    }
  }
```

### DIFF
--- a/modules/chat/src/Chat.ts
+++ b/modules/chat/src/Chat.ts
@@ -1,6 +1,6 @@
 import {
   ManagedModule,
-  DatabaseProvider,
+  DatabaseProvider, ConfigController, Email,
 } from '@conduitplatform/grpc-sdk';
 
 import AppConfigSchema from './config';
@@ -29,6 +29,7 @@ export default class Chat extends ManagedModule {
   private adminRouter: AdminHandlers;
   private userRouter: ChatRoutes;
   private database: DatabaseProvider;
+  private emailModule: Email;
 
   constructor() {
     super('chat');
@@ -40,10 +41,16 @@ export default class Chat extends ManagedModule {
   }
 
   async onConfig() {
+    const config = await ConfigController.getInstance().config;
+    if (config.sendInvitations.enabled) {
+      if (config.sendInvitations.send_email)
+        await this.grpcSdk.waitForExistence('email');
+      this.emailModule = this.grpcSdk.emailProvider!;
+    }
     if (!this.isRunning) {
       await this.registerSchemas();
       this.adminRouter = new AdminHandlers(this.grpcServer, this.grpcSdk);
-      this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk);
+      this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk,this.emailModule);
       this.isRunning = true;
     }
     await this.userRouter.registerRoutes();
--- a/modules/chat/src/Chat.ts
+++ b/modules/chat/src/Chat.ts
@@ -42,16 +42,16 @@ export default class Chat extends ManagedModule {
 
   async onConfig() {
     const config = await ConfigController.getInstance().config;
-    if (config.sendInvitations.enabled) {
-      if (config.sendInvitations.send_email)
-        await this.grpcSdk.waitForExistence('email');
-      this.emailModule = this.grpcSdk.emailProvider!;
-    }
     if (!this.isRunning) {
       await this.registerSchemas();
       this.adminRouter = new AdminHandlers(this.grpcServer, this.grpcSdk);
       this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk,this.emailModule);
       this.isRunning = true;
+    }
+    if (config.sendInvitations.enabled) {
+      if (config.sendInvitations.send_email)
+        await this.grpcSdk.waitForExistence('email');
+      this.emailModule = this.grpcSdk.emailProvider!;
     }
     await this.userRouter.registerRoutes();
   }
--- a/modules/chat/src/Chat.ts
+++ b/modules/chat/src/Chat.ts
@@ -45,7 +45,7 @@ export default class Chat extends ManagedModule {
     if (!this.isRunning) {
       await this.registerSchemas();
       this.adminRouter = new AdminHandlers(this.grpcServer, this.grpcSdk);
-      this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk,this.emailModule!);
+      this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk);
       this.isRunning = true;
     }
     if (config.explicit_room_joins.send_email)
--- a/modules/chat/src/Chat.ts
+++ b/modules/chat/src/Chat.ts
@@ -29,7 +29,6 @@ export default class Chat extends ManagedModule {
   private adminRouter: AdminHandlers;
   private userRouter: ChatRoutes;
   private database: DatabaseProvider;
-  private emailModule?: Email;
 
   constructor() {
     super('chat');
@@ -48,10 +47,8 @@ export default class Chat extends ManagedModule {
       this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk);
       this.isRunning = true;
     }
-    if (config.explicit_room_joins.send_email)
+    if (config.explicit_room_joins.enabled && config.explicit_room_joins.send_email)
       await this.grpcSdk.waitForExistence('email');
-    this.emailModule = this.grpcSdk.emailProvider!;
-
     await this.userRouter.registerRoutes();
   }
 
--- a/modules/chat/src/Chat.ts
+++ b/modules/chat/src/Chat.ts
@@ -29,7 +29,7 @@ export default class Chat extends ManagedModule {
   private adminRouter: AdminHandlers;
   private userRouter: ChatRoutes;
   private database: DatabaseProvider;
-  private emailModule: Email;
+  private emailModule?: Email;
 
   constructor() {
     super('chat');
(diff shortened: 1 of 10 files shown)


modules/chat/src/routes/routes.ts
```diff
@@ -72,23 +92,84 @@ export class ChatRoutes {
...
+    if (config.sendInvitations.enabled) {
+      const serverConfig = await this.grpcSdk.config.getServerConfig();
+      const url = serverConfig.url;
+      for (const invitedUser of usersToBeAdded) {
+        const invitationsCount = await InvitationToken.getInstance().countDocuments({
+          room: roomId,
+          receiver: invitedUser._id,
+          sender: user._id,
+        });
+        if (room.participants.includes(invitedUser._id)) {
```
kon14, line 111: No need to complete the loop.
Both checks should immediately throw an error after finding a single existing member or pre-invited user to avoid wasting cycles on a failed request.

eg `throw new GrpcError(status.INVALID_ARGUMENT, 'users array contains existing member or invited user ids');`

modules/chat/src/routes/routes.ts
```diff
@@ -72,23 +92,84 @@ export class ChatRoutes {
...
+          });
+        }
+        retMessage.push({
+          receiver: invitedUser._id,
+          message: 'Invitation sent',
+        });
+      }
+      return retMessage;
+    } else {
+      room.participants = Array.from(new Set([...room.participants, ...users]));
```
kon14, line 155: This case could also use a member exists check throwing an error for already included users:
`throw new GrpcError(status.INVALID_ARGUMENT, 'users array contains existing member ids');`

kon14 (changes requested):

modules/chat/src/routes/routes.ts
```diff
@@ -343,6 +424,140 @@ export class ChatRoutes {
...
+    });
+    if (isNil(invitationTokenDoc)) {
+      throw new GrpcError(status.NOT_FOUND, 'Invitation not valid');
+    }
+    let message;
+    let receiver = user._id;
+    const accepted = (answer === 'accept');
+    const chatRoom = await ChatRoom.getInstance().findOne({ _id: invitationTokenDoc.room });
+    if (isNil(chatRoom)) {
+      throw new GrpcError(status.NOT_FOUND, 'Room not exists');
```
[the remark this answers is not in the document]
sdimitris: xd

modules/chat/src/Chat.ts
```diff
@@ -29,6 +29,7 @@ export default class Chat extends ManagedModule {
   private adminRouter: AdminHandlers;
   private userRouter: ChatRoutes;
   private database: DatabaseProvider;
+  private emailModule: Email;
```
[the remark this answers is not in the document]
kon14: Actually we shouldn't have that there in the first place. Client can be obtained through `grpc-sdk`.

modules/chat/src/routes/InvitationRoutes.ts
```diff
@@ -0,0 +1,243 @@
...
+    });
+    if (isNil(invitationTokenDoc)) {
+      throw new GrpcError(status.NOT_FOUND, 'Invitation not valid');
+    }
+    let message;
+    let receiver = user._id;
+    const accepted = (answer === 'accept');
+    const chatRoom = await ChatRoom.getInstance().findOne({ _id: invitationTokenDoc.room });
+    if (isNil(chatRoom)) {
+      throw new GrpcError(status.NOT_FOUND, 'Room not exist');
```
kon14, line 123: Still here. Error message should be `Room does not exist`.
Ideally keep this consistent across all routes (answerInvitationFromHook responds with `Chat room does not exist`)

modules/chat/src/Chat.ts
```diff
@@ -40,12 +41,17 @@ export default class Chat extends ManagedModule {
...
 
   async onConfig() {
+    const config = await ConfigController.getInstance().config;
     if (!this.isRunning) {
       await this.registerSchemas();
       this.adminRouter = new AdminHandlers(this.grpcServer, this.grpcSdk);
       this.userRouter = new ChatRoutes(this.grpcServer, this.grpcSdk);
       this.isRunning = true;
     }
+    if (config.explicit_room_joins.send_email)
```
kon14, line 51: This should be `if (config.explicit_room_joins.enabled && config.explicit_room_joins.send_email`.
Same with any other check that ends up awaiting email or sending invitation emails.

If we only have `config.explicit_room_joins.enabled === true` we merely create the invitations.
`send_email === true` is ignored while `config.explicit_room_joins.enabled === false`.

modules/chat/src/routes/InvitationRoutes.ts
```diff
@@ -0,0 +1,243 @@
...
+      undefined,
+      'receiver',
+    )
+      .catch((e: Error) => {
+        throw new GrpcError(status.INTERNAL, e.message);
+      });
+    const count = invitations.length;
+    return { invitations, count };
+  }
+}
```
kon14, line 243: Add a new line at the end of file.

modules/chat/src/routes/InvitationRoutes.ts
```diff
@@ -0,0 +1,243 @@
...
+          answer: ConduitString.Required,
+          invitationToken: ConduitString.Required,
+        },
+      },
+      new ConduitRouteReturnDefinition('InvitationResponse', 'String'),
+      this.answerInvitationFromHook.bind(this),
+    );
+    this.routingManager.route(
+      {
+        path: '/invitation/:answer/:id',
```
kon14, line 49: `/invitation/:answer/:id` -> `/invitations/:answer/:id`

modules/chat/src/routes/InvitationRoutes.ts
```diff
@@ -0,0 +1,243 @@
...
+      new ConduitRouteReturnDefinition('SentInvitationsResponse', {
+        invitations: [InvitationToken.getInstance().fields],
+        count: ConduitNumber.Required,
+
+      }),
+      this.sentInvitations.bind(this),
+    );
+    this.routingManager.route(
+      {
+        path: '/invitation/cancel/:id',
```
kon14, line 95: `/invitation/cancel/:id` -> `/invitations/cancel/:id`

modules/chat/src/routes/InvitationRoutes.ts
```diff
@@ -0,0 +1,243 @@
...
+      })
+      .catch((e: Error) => {
+        console.error('Internal error while registering email templates');
+      });
+  }
+
+  declareRoutes() {
+    this.routingManager.route(
+      {
+        path: '/hook/:answer/:invitationToken',
```
kon14, line 37: Make this `/hook/invitations/:answer/:invitationToken` (and update uri in mail templates etc).

modules/chat/src/utils/index.ts
```diff
@@ -23,6 +25,64 @@ export async function validateUsersInput(grpcSdk: ConduitGrpcSdk, users: any[])
       });
     }
   }
+  return usersToBeAdded;
+}
+
+export async function sendInvitations(users: any, sender: User, room: ChatRoom, url: string, config: any, grpcSdk: ConduitGrpcSdk) {
```
kon14, line 31: Add type for users.
Pass a `sendEmail` boolean arg instead of the whole module config.

modules/chat/src/routes/routes.ts
kon14: ^

modules/chat/src/utils/index.ts
```diff
@@ -23,6 +25,64 @@ export async function validateUsersInput(grpcSdk: ConduitGrpcSdk, users: any[])
...
+      room: roomId,
+      receiver: invitedUser._id,
+      sender: sender._id,
+    });
+    if (room.participants.includes(invitedUser._id)) {
+      retMessage.push({
+        receiver: invitedUser._id,
+        message: 'Already member',
+      });
+      continue;
```
kon14, line 46: What I meant to say is, there's no need to complete the loop nor move on to the next one.
Both this and the invitationsCount check below should immediately throw an error.

`throw new GrpcError(status.INVALID_ARGUMENT, 'users array contains existing member or invited user ids');`

We could technically find all the possible problems with the query and return the appropriate error message collection, but we don't do that anywhere else and it's mostly a waste of cycles as the request should fail anyway.

kon14 (changes requested):

modules/chat/src/routes/routes.ts
```diff
@@ -466,7 +686,80 @@ export class ChatRoutes {
         this.deleteMessage.bind(this),
       );
     }
+    if (config.sendInvitations.enabled) {
+      // TODO Check if email is active
```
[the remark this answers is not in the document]
kon14: Perform the check up here, then [pass a boolean to sendInvitations()](https://github.com/ConduitPlatform/Conduit/pull/119#discussion_r852426384).

This is technically handled in the new util function, but we shouldn't be passing the entire config object there (and if we really needed the config we could be getting it through the ConfigController singleton anway).

modules/chat/src/routes/InvitationRoutes.ts
sdimitris: Xd

kkopanidis (approved):

modules/chat/src/routes/InvitationRoutes.ts
kon14: This is still wrong.

modules/chat/src/routes/InvitationRoutes.ts
kon14: This is still wrong.

kon14 (changes requested):
This is still wrong.
`/invitation/cancel/:id` -> `/invitations/cancel/:id`

Issue #112, opened 2022-04-05
[BUG] Race condition between Config module and module configuration requests


kon14:
**Describe the bug**

The `Config` module currently expects any configurable modules being brought up, or at the very least attempt communication, after `Database` so that the `Config` schema can be registered and the module service becoming functional before any incoming requests make their way through gRPC.

This is especially a problem for modules expecting to migrate their configuration early in their initialization.

**To Reproduce**
1. Bring up `Core`
2. Start `Authentication`
3. Start `Database`
4. Notice errors (eg `No database instance provided!` as a result of uninitialized `Config` model singleton)

**Expected behavior**
Config should probably be caching early configuration requests and performing them once ready(?)
Otherwise, Conduit's initialization will have to be reworked so that `Config` is becomes serviceable before modules get a chance to communicate with it.

kon14:
This is going to require certain changes to Conduit's initialization, proper gRPC health checks (#29) are going to be required for grpc-sdk to handle this efficiently.

I'm currently looking into how best to implement this.
Database introspection is also an important consideration.

PR #126, opened 2022-04-21
feat: implement gRPC health check protocol

### LINKED ISSUE #29: [FEAT] Support proper Health Checking
Currently health checking is being done by this simple process:
https://github.com/ConduitPlatform/Conduit/blob/7f4bcf9d98e1d54f020aedd87c88eb01655db8d6/packages/config/src/index.ts#L474-L502

It works well enough, but it's not common practice and it simply checks if the grpc server is online. We're getting no proper health statuses and we're using an http client to check a grpc server which isn't really that great.

The modules should implement (through the grpc-sdk) the Health Checking protocol of grpc as defined here:
https://github.com/grpc/grpc/blob/master/doc/health-checking.md

The config module should then be able to use the Watch rpc stream to get health status updates or at the very list make individual requests (as it's happening now but through grpc) to get better sense of the actual service status. 

It'll also be more in line with common practices around grpc

kon14:
This PR provides support for gRPC health checking for Core and modules based on the [GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md).

It also improves general startup behavior as modules may now be started in any random order, including before `Core` itself, and still properly connect as soon as `Core` is online and ready to serve them.
All modules wait on `Core`'s gRPC service being brought up and reporting as `SERVING`, with the exception of `Database` that registers on `UNKNOWN`.

Closes #29 and #112.

Also fixes inactive/offline modules not being removed from `Config`'s registered modules.

- [X] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [X] Yes
- [ ] No

This PR does not affect the APIs so applications relying on Conduit are not actually affected.
Modules simply require a rebuld using the latest grpc-sdk.


libraries/grpc-sdk/src/classes/ConduitServiceModule.ts
```diff
@@ -1,9 +1,11 @@
 import ConduitGrpcSdk, {
   GrpcServer,
+  HealthCheckStatus,
 } from '..';
+import path from 'path';
 
 export abstract class ConduitServiceModule {
```
kkopanidis, line 7: Instead of simply basing further checks on overrides, functions can be exposed so that the module can do something like grpcSdk.updateHealth(Healthy);

libraries/grpc-sdk/src/classes/ModuleManager.ts
```diff
@@ -22,38 +22,41 @@ export class ModuleManager {
...
-        process.exit(-1);
-      })
-      .then(() => {
-        return this.postRegisterLifecycle();
-      })
-      .catch((err: Error) => {
-        console.log('Failed to activate module');
-        console.error(err);
-      });
+    this.grpcSdk.initialize().then(() => {
```
kkopanidis, line 25: I reeeally dislike seeing these then blocks, isn't there a better way to write this?

libraries/grpc-sdk/src/classes/ModuleManager.ts
```diff
@@ -22,38 +22,41 @@ export class ModuleManager {
...
+    });
   }
 
   private async preRegisterLifecycle(): Promise<void> {
+    await this.module.createGrpcServer(this.servicePort);
     await this.module.preServerStart();
-    await this.module.startGrpcServer(this.servicePort);
-    await this.module.onServerStart()
+    await this.module.startGrpcServer();
+    await this.module.onServerStart();
```
kkopanidis, line 54: I'm thinking that the bus should probably be initialized before the server start. Starting the server assumes we can actually serve stuff. But in multi-instance cases, the bus is needed very early on.

libraries/grpc-sdk/src/modules/config/index.ts
```diff
@@ -103,14 +108,14 @@ export class Config extends ConduitModule<typeof ConfigDefinition> {
     };
     const self = this;
     this.client!.moduleHealthProbe(request)
```
kkopanidis, line 110: This should typically be replaced by the health checks. Maybe the entire complexity of this (including the service/clientNames) could be solved by merging the core and config packages. Since the Core only bootstraps stuff but it's very dependent on config. So merging Config and Core would be a reasonable move. This also brings us closer to also enabling the optional splitting (during deployment) of core/config and the routers, but I'm getting off-topic here.

modules/database/src/Database.ts
```diff
@@ -54,6 +62,16 @@ export default class DatabaseModule extends ManagedModule {
   private adminRouter: AdminHandlers;
   private userRouter: DatabaseRoutes;
   private readonly _activeAdapter: DatabaseAdapter<MongooseSchema | SequelizeSchema>;
+  private readonly events: EventEmitter = new EventEmitter();
+  private _serviceHealthState: HealthCheckStatus = HealthCheckStatus.UNKNOWN;
+
+  private set serviceHealthState(state: HealthCheckStatus) {
```
kkopanidis, line 68: This could be simplified by the suggestion I added in a previous comment. It'll also remove the event emitter which is costly performance-wise in NodeJS.

packages/config/src/index.ts
```diff
@@ -55,39 +53,28 @@ export default class ConfigManager implements IConfigManager {
     this.sdk
```
kkopanidis, line 53: Either change this to an "await" only approach, or revert it. Mixing awaits and then/catch Promises is very ugly.

packages/config/src/index.ts
```diff
@@ -99,7 +86,7 @@ export default class ConfigManager implements IConfigManager {
               messageParsed.name,
               messageParsed.url,
               messageParsed.instance,
-            );
+            ).then();
```
kkopanidis, line 89: Just no.

kkopanidis (changes requested):

libraries/grpc-sdk/src/classes/ConduitModule.ts
```diff
@@ -1,24 +1,25 @@
...
-  // private readonly _interceptors: any[];
-
-  constructor(name: string, url: string) {
-    this._name = name;
-    this._url = url;
-    // this._interceptors = [
-    //   getModuleNameInterceptor(this._name),
-    // ];
+  constructor(clientName: string, serviceName: string, serviceUrl: string) {
+    this._clientName = clientName;
```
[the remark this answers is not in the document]
kon14: `clientName` is not `serviceName` for any module as the former refers to the module initializing a gRPC client.
I could rename these to `clientModuleName` and `serverModuleName` for clarification.

Both the check and watch functions accept a service variable, so the latter could just emit an event based on that, but service names are not necessarily unique across modules so that is probably unwise.
Ideally the watcher function should emit an event formatted with both the server module name and the service name.

The watcher function could accept an event name instead, that way we can get rid of it while at the same time make it so that module authors know what event name to listen for.
If we choose not to, we should probably at the very least return the formatted event name so module authors know what to listen for without having to read how `ConduitModule` works.

libraries/grpc-sdk/src/classes/ModuleManager.ts
kon14: I guess as `ModuleManager.start()` is the last thing to be called before module initialization takes over in `index.ts` files we could just turn the whole thing async without even requiring top-level await.

libraries/grpc-sdk/src/classes/ConduitServiceModule.ts
kon14: Fine by me, but we'd also be adding multi-service complexity to `ConduitServiceModule`, as modules exposing multiple services would need a way to update/check/watch these individually.

The class should be aware of the module name, supported service names, keep track of each service's state and handle cases where no service name is provided in a health check/watch request.
The last bit is not entirely obvious. Usually it'd make sense to respond <=non-serving as long as a any service is not serving, but what about if a module would still like to respond as `SERVING` when a non-critical service is offline etc?

I guess it could still just override the function at this point if anything other the normally expected behavior is desired.

libraries/grpc-sdk/src/index.ts
```diff
@@ -50,12 +54,37 @@ export default class ConduitGrpcSdk {
...
     }
     this.serverUrl = serverUrl;
-    this._config = new Config(this.name, this.serverUrl);
-    this._admin = new Admin(this.name, this.serverUrl);
-    this._router = new Router(this.name, this.serverUrl);
-    this.initializeModules().then(() => {
-    });
+  }
+
+  async initialize() {
```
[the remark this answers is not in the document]
kon14: This is not core/db-specific.
Modules now wait for core to be online and serving before moving on with their own initialization.
That way we can launch modules and prevent them from crashing regardless of which order they're brought up in.

For example we could launch Authentication before Core is capable of serving anyone or even started in the first place.
Imo this would make even more sense post-Database-introspection, where Database could take a while before being initially registered for anyone connecting to a populated db.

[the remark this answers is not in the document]
kon14: The while loop part isn't ideal either, but using a core health watch callback there would probably be messy as far as module initialization is concerned.

Might still end having to use a watch anyway to prevent core from getting bombarded by requests though.

modules/database/src/Database.ts
kon14: I moved this logit to `ConduitServiceModule`, but we can't just get rid of the emitter, unless we keep track of watch calls on the class and write back data on every state change directly inside `updateHealth()`.

kkopanidis (approved):

Issue #100, opened 2022-03-30
[BUG] Default sequelize schema ids conflict with primary keys


Michael-Vol:
**Describe the bug**
Currently, the sequelize schema adds a default uuid (_id) as a primary key to the model. In case the user provides a database for introspection and the tables have already a primary key, there will be a conflict between that primary key and the _id we are trying to add to the schema.
We need to keep both ids though, since Conduit uses the _id for any queries related to that model.
One possible solution would be to change the _id from primary key to virtual, however sequelize currently doesn't support converting the virtual datatype into the sql equivalent (see https://github.com/sequelize/sequelize/issues/12451).

**To Reproduce**
1. Start the database module with an existing database connection in the envs
2. Notice that the database won't sync because of the virtual field is not defined.

**Expected Behavior**
The database module should introspect the db by keeping both it's already defined primary key and adding the conduit _id used for future queries.

PR #128, opened 2022-04-26
feat(database): DB introspection

### LINKED ISSUE #77: [FEAT] Database Introspection
Currently if a user wants to connect Conduit into an existing database, they will have to import all existing schemas one by one. A process which will be tedious and error-prone, with possible destructive consequences on SQL DBs

We need to add a feature, that will also require new UI in the admin panel, which once connected to a database Conduit will provide the user with the option to introspect the DB. For SQL dbs this is fairly straightforward since the schema is pretty-much already there, but for MongoDB we'll need to do document scans to understand the schema of each collection. 

**Additional context**
Some things to consider:
- This process absolutely needs the admin UI, so that the user can confirm what Conduit detected and wants to import so that no migrations get fired incorrectly.
- For MongoDB we may even need to support multiple schemas that target the same collection. Code-wise this is already supported, but on the admin UI it is assumed that a schema and a collection have 1-1 relations
- Indexes are not properly supported but will be required for introspection since SQL uses them for foreign_keys (I think).
- Foreign key support is kinda sketchy as far as I can remember so we definitely need to look into that
- The database module will probably need some sort of change, because once online it will try to register clientid/secret schemas, the declaredSchema etc. We have virtually 2 options. 1 make sure that our schema naming is super unique, 2 check for schema existence before trying to register. And if the schema exists in the actual database but not in our declared schemas, we should print an error that an introspection needs to be made first. This still leaves us with the issue of ex. AccessToken schema from Authentication were if it already exists we cannot use it and the modules should't handle fallback scenarios. So probably the database will either need to make the schema collection name unique, or some other approach that I can't think of.

Michael-Vol:
This PR provides support for database introspection, allowing the user to connect an existing database into Conduit.
- The process is completely synchronous, the database module introspects all existing tables/collections before registering the routes.
- An sql schema name can be provided as an env variable(SQL_SCHEMA), if the user wants to add the tables from a specified schema instead of the default (public).
- A reintrospection route has also been added, in case the user has added additional non-conduit schemas to the database.

Closes #77 and #100.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No


kkopanidis (changes requested):

modules/database/src/adapters/DatabaseAdapter.ts
```diff
@@ -63,7 +74,7 @@ export abstract class DatabaseAdapter<T extends SchemaAdapter<any>> {
   ): { model: SchemaAdapter<any>; relations: any };
 
   fixDatabaseSchemaOwnership(schema: ConduitSchema) {
-    const dbSchemas = ['CustomEndpoints'];
+    const dbSchemas = ['CustomEndpoints','_PendingSchemas'];
```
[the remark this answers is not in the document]
kon14: @kkopanidis It is required in order to call `createSchemaFromAdapter()` from `Database` (that is to say, without going through gRPC), as otherwise the `ConduitSchema` would get an undefined `ownerModule` and fail the `getSchemaOwnership()` check.

We could also explicitly set the object owner post schema instantiation for database models I guess?

kon14:
It should be noted how this PR comes with a couple limitations/considerations:

1. Multiple Postgres search path "Schemas" are not supported (specify `SQL_SCHEMA` env, defaults to `public`).
2. SQL primary keys are not based on indexes, but rather specified on the fields (composite primary keys specify multiple primary keys).

Working around the above Conduit-specific assumptions requires further changes to ConduitSchema and across Conduit in general.

modules/database/src/adapters/sequelize-adapter/index.ts
```diff
@@ -1,11 +1,21 @@
...
+import { SequelizeAuto, TableData } from 'sequelize-auto';
+import {
+  sqlSchemaConverter,
+  INITIAL_DB_SCHEMAS,
+} from '../../introspection/sequelize/utils';
+import { isNil } from 'lodash';
+import { isMatch } from 'lodash';
+import { MultiDocQuery } from '../../interfaces';
+
+const sqlSchemaName = process.env.SQL_SCHEMA ?? 'public';
```
kon14, line 18: Consider specifying Sequelize's `options.searchPath` so that we don't need to explicitly specify this everywhere as that could potentially be a root cause for unexpected bugs down the line as soon as we forget about it.

[Link to Sequelize docs](https://sequelize.org/api/v6/class/src/model.js~model) (options.searchPath)

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -30,6 +38,7 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
     this.registeredSchemas = new Map();
     this.connectionString = connectionString;
     this.mongoose = new Mongoose();
+    this.mongoose.pluralize(null);
```
kon14, line 41: Conduit should explicitly pluralize schema names consistently across both Mongoose and Sequelize.

This is going to require a **migration** for existing SQL deployments as their current system schema names are going to differ. Notably, `_DeclaredSchema !== _DeclaredSchemas` which will be result in the collection getting recreated ootb.
We could probably(?) avoid migrating declared schema entries by checking against the collection contents or something if we don't wish to migrate everything.

modules/database/src/adapters/DatabaseAdapter.ts
```diff
@@ -8,6 +8,17 @@ export abstract class DatabaseAdapter<T extends SchemaAdapter<any>> {
   registeredSchemas: Map<string, ConduitSchema>;
   models?: { [name: string]: T };
 
+  /**
+   * Checks if the database adapter includes the declaredSchema model
```
kon14, line 12: Minor nitpicking, but explain what the function does rather than how it does it.
It's a lie waiting to happen.

kon14 (changes requested):

modules/database/src/adapters/sequelize-adapter/index.ts
Michael-Vol: We could add the searchPath option however it cannot be provided to the sequelize constructor, meaning that we would need to explicitly add it to every CRUD operation on the models themselves rather than the sequelize instance.

modules/database/src/Database.ts
```diff
@@ -80,18 +82,26 @@ export default class DatabaseModule extends ManagedModule {
   }
 
   async onServerStart() {
-    await this._activeAdapter.createSchemaFromAdapter(models.DeclaredSchema);
-    // Introspection takes place
+    const isConduitDB = await this._activeAdapter.isConduitDB();
+    if (isConduitDB) {
+      await this._activeAdapter.createSchemaFromAdapter(models.DeclaredSchema);
+    } else {
+      await this.introspectDB();
```
kon14, line 89: The reason why this check was added was so that we could create `DeclaredSchema` **after** schema introspection to prevent `isConduitDB()` from failing on an interrupted introspection.
Tbh this still wouldn't completely prevent a half-way introspection as it could crash after the schemas have been introspected and before they have been stored.
After adding `PendingSchemas`, `introspectDB()` ended up creating it anyway, therefore: 

- `DeclaredSchema` should always be created at the top
- `isConduitDB()` should be smarter (eg check if a core schema exists as `Core` only registers these after the initial introspection is over), 
- If `!isConduitDB()`, run `introspectDB()`, creating `PendingSchemas` etc

modules/database/src/Database.ts
```diff
@@ -541,4 +547,27 @@ export default class DatabaseModule extends ManagedModule {
       });
     }
   }
+
+  private async introspectDB() {
```
kon14, line 551: Convert to proper camelCase.

modules/database/src/adapters/DatabaseAdapter.ts
```diff
@@ -8,6 +8,16 @@ export abstract class DatabaseAdapter<T extends SchemaAdapter<any>> {
   registeredSchemas: Map<string, ConduitSchema>;
   models?: { [name: string]: T };
 
+  /**
+   * Checks if the database has already been connected with Conduit
+   */
+  abstract isConduitDB(): Promise<boolean>;
```
kon14, line 14: Convert to proper camelCase.

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -75,6 +84,117 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
       });
   }
 
+  async isConduitDB(): Promise<boolean> {
```
kon14, line 87: Convert to proper camelCase.

modules/database/src/adapters/DatabaseAdapter.ts
kon14: Check whether any of `Core`'s schemas exist instead.
Introspection could fail half-way, whereas `Core` only gets a chance to register any of its schemas after the initial introspection is finalized.

kon14 (changes requested):

PR #150, opened 2022-05-12
feat: reactive module dependency handling

kon14:
This PR makes introduces the following changes:
1. Modules do not respond as `SERVING` when they're configured as inactive
2. Any modules depending on other modules actually check whether said modules are in a `SERVING` state instead of assuming they're serviceable as long as they are online.
3. Dependent modules react to any dependency serving state changes, meaning they can react to live changes in the distributed environment and handle their own functionality according to what's currently available.

Modules **hard-depending** on offline/non-serving modules (eg `Chat -> Authentication`) won't be brought up until said conditions are met, while also updating their own state to `NOT_SERVING` as soon as they're notified about their dependencies being unavailable.

Models **soft-depending** on offline/non-serving modules for secondary features (eg `Chat -> Email` based on configuration) will disable their optional features whenever said modules are unavailable and turn them back on when their conditions are met.

Two functions have been added to `grpc-sdk`:
- `monitorModule()`: Registering a new callback to handle state changes for a specific module.
Explicitly executes the callback once against checked module state so that handler code can be executed on dependent module startup.
- `unmonitorModule()`: Unregistering any event listeners on target module's state change events.
Required for config-based dependencies that may get disabled without a restart.

Note on 2FA:
`Authentication` deployments enabling 2FA support (deployment config + opt-in by user) won't disable 2FA while `SMS` is offline, meaning 2FA users are unable to authenticate themselves while all instances of `SMS` are down.

Note on disabling routes:
Live removal of previously registered routes is not currently supported by Conduit.
As such, module reconfiguration and dependency status updates are incapable of removing dependency-based routes.

Additionally, any `waitForExistence('database')` calls have been removed as, as of #126, `Database` is always registered before any other module can be brought online and ending up with no db instance is currently not recoverable down the line anyway.

This PR also includes fixes for `Config` (state update and recovery) and `grpc-sdk` (`updateHealth()` no longer broadcasts a state update event whenever a new state is identical to the old one).

- [X] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No


libraries/grpc-sdk/src/index.ts
```diff
@@ -21,12 +21,14 @@ import { ConduitModule } from './classes/ConduitModule';
 import { Client } from 'nice-grpc';
 import { status } from '@grpc/grpc-js';
 import { sleep } from './utilities';
-import { HealthDefinition } from './protoUtils/grpc_health_check';
+import { HealthDefinition, HealthCheckResponse_ServingStatus } from './protoUtils/grpc_health_check';
```
kkopanidis, line 24: protoUtils is a generated folder, isn't there a better way to do this?

libraries/grpc-sdk/src/index.ts
```diff
@@ -219,20 +224,36 @@ export default class ConduitGrpcSdk {
     this.config.watchModules().then();
     emitter.on('serving-modules-update', (modules: any) => {
       Object.keys(this._modules).forEach((r) => {
```
kkopanidis, line 226: Can we look into merging these 2 for loops? They seem to be mergeable, and I definitely don't remember why I had this split.

libraries/grpc-sdk/src/index.ts
```diff
@@ -219,20 +224,36 @@ export default class ConduitGrpcSdk {
...
+          }
+          this.createModuleClient(m.moduleName, m.url);
+          emitter.emit(`module-connection-update:${m.moduleName}`, { serving: true });
         }
-        this.createModuleClient(m.moduleName, m.url);
       });
     });
   }
 
+  async monitorModule(moduleName: string, callback: (args: { serving: boolean }) => void) {
```
kkopanidis, line 245: simplify the callback. No reason for an object.

modules/authentication/src/Authentication.ts
```diff
@@ -68,12 +69,32 @@ export default class Authentication extends ManagedModule<Config> {
...
   async onConfig() {
-    if (!this.isRunning) {
+    const config = ConfigController.getInstance().config;
+    if (!config.active) {
+      this.updateHealth(HealthCheckStatus.NOT_SERVING);
+    } else {
+      let refreshedOnce = false;
       await this.registerSchemas();
+      if (config.local.verification.send_email) {
+        await this.grpcSdk.waitForExistence('email');
```
kkopanidis, line 79: wait for existence should be included in the email. I'm not saying that grpc-sdk shouldn't provide the waitForExistence function, but the monitor should use it, since it's to be expected.

modules/authentication/src/handlers/local.ts
```diff
@@ -586,7 +591,7 @@ export class LocalHandlers {
   private async initDbAndEmail() {
     const config = ConfigController.getInstance().config;
 
-    if (config.local.verification.send_email) {
+    if (this.sendEmail) {
       await this.grpcSdk.config.moduleExists('email');
```
kkopanidis, line 595: why await? If sendEmail is true you already know the email module is up.

modules/chat/src/Chat.ts
```diff
@@ -26,32 +27,74 @@ export default class Chat extends ManagedModule<Config> {
...
 
   constructor() {
     super('chat');
+    this.updateHealth(HealthCheckStatus.UNKNOWN, true);
   }
 
   async onServerStart() {
-    await this.grpcSdk.waitForExistence('database');
     this.database = this.grpcSdk.databaseProvider!;
+    await this.grpcSdk.waitForExistence('authentication');
```
kkopanidis, line 53: same comment as in auth

modules/forms/src/Forms.ts
```diff
@@ -24,12 +29,19 @@ export default class Forms extends ManagedModule<Config> {
...
   constructor() {
     super('forms');
+    this.updateHealth(HealthCheckStatus.UNKNOWN, true);
   }
 
   async onServerStart() {
-    await this.grpcSdk.waitForExistence('database');
-    await this.grpcSdk.waitForExistence('email');
     this.database = this.grpcSdk.databaseProvider!;
+    await this.grpcSdk.waitForExistence('email');
```
kkopanidis, line 37: same

modules/push-notifications/src/PushNotifications.ts
```diff
@@ -41,16 +42,27 @@ export default class PushNotifications extends ManagedModule<Config> {
 
   constructor() {
     super('pushNotifications');
+    this.updateHealth(HealthCheckStatus.UNKNOWN, true);
   }
 
   async onServerStart() {
-    await this.grpcSdk.waitForExistence('database');
     this.database = this.grpcSdk.databaseProvider!;
+    await this.grpcSdk.waitForExistence('authentication');
```
kkopanidis, line 50: same

kkopanidis (changes requested):

libraries/grpc-sdk/src/index.ts
kon14: Config won't send over any removed modules, as such we would still have to diff the arrays and `closeConnection()` on any removed modules.

libraries/grpc-sdk/src/index.ts
```diff
@@ -219,20 +224,46 @@ export default class ConduitGrpcSdk {
...
     });
   }
 
+  async monitorModule(
+    moduleName: string,
+    callback: (serving: boolean) => void,
+    wait: boolean = true,
+  ) {
+    if (wait) await this.waitForExistence(moduleName);
+    this._modules['chat']?.healthClient?.check({})
```
kkopanidis, line 251: whooops

kkopanidis (approved):

PR #208, opened 2022-06-15
fix(security): security client aliases are not unique

kon14:
This PR enforces uniqueness across security client aliases.
We already have a notes field for non-unique descriptions anyway.

Security clients now have a db-level uniqueness constraint, and as such, they are also required on a db-level.
Explicitly passing in an alias on security client creation is still optional as it is otherwise implicitly generated based on the platform type, client id and, in case of web clients, the domain used.

~~This is not a breaking change as the feature is not formally released in a tagged non-release-candidate version.~~
This PR **does** introduce a breaking change related to `createClient` and `updateClient` admin route return types.
Basically, we now return `_id` instead of `id` (plus the remaining missing model fields) as we generally do everywhere else.
I am not marking the commit itself as a Breaking Change as we're already in a release candidate for v0.13, but we should include this in the final release notes.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [X] Yes
- [ ] No

### DIFF
--- a/packages/security/src/admin/routes/CreateSecurityClient.route.ts
+++ b/packages/security/src/admin/routes/CreateSecurityClient.route.ts
@@ -39,6 +39,23 @@ export function getCreateSecurityClientRoute() {
       if (!Object.values(PlatformTypesEnum).includes(platform)) {
         throw new ConduitError('INVALID_ARGUMENTS', 400, 'Platform not supported');
       }
+      if (alias) {
+        const existingClient = await Client.getInstance().findOne({ alias });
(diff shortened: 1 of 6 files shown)


kon14:
Pushed a bunch of fixes around alias uniqueness (db constraint, auto-generated if not explicitly provided), model class fields and handlers.

kon14:
Triggering a CodeFactor check re-run post #210 merge.

PR #223, opened 2022-07-06
feat(database): add reconnect attempts to sequelize

Michael-Vol:
This PR adds support for multiple connection attempts in the Sequelize adapter in case connection fails.
- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/modules/database/src/adapters/sequelize-adapter/index.ts
+++ b/modules/database/src/adapters/sequelize-adapter/index.ts
@@ -29,22 +29,7 @@ export class SequelizeAdapter extends DatabaseAdapter<SequelizeSchema> {
   }
 
   async connect() {
-    let error;
-    for (let i = 0; i < 10; i++) {
-      try {
-        console.log(`${i} attempts`);
-        return (this.sequelize = new Sequelize(this.connectionUri, { logging: false }));
-      } catch (err) {
-        await sleep(200);
-        error = err;
-        continue;
-      }
-      error = undefined;
-    }
-    if (error) {
-      ConduitGrpcSdk.Logger.error(error);
-      throw new GrpcError(status.INTERNAL, 'Connection with DB not possible');
-    }
+    return (this.sequelize = await new Sequelize(this.connectionUri, { logging: false }));
   }
 
   async retrieveForeignSchemas(): Promise<void> {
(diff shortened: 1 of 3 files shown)


modules/database/src/adapters/sequelize-adapter/index.ts
```diff
@@ -27,8 +28,8 @@ export class SequelizeAdapter extends DatabaseAdapter<SequelizeSchema> {
     this.connectionUri = connectionUri;
   }
 
-  connect() {
-    this.sequelize = new Sequelize(this.connectionUri, { logging: false });
+  async connect() {
+    return (this.sequelize = await new Sequelize(this.connectionUri, { logging: false }));
```
kon14, line 32: `connect()`'s signature returns void, no need to return the instance here.

modules/database/src/adapters/sequelize-adapter/index.ts
```diff
@@ -255,15 +256,24 @@ export class SequelizeAdapter extends DatabaseAdapter<SequelizeSchema> {
   }
 
   async ensureConnected() {
-    return this.sequelize
-      .authenticate()
-      .then(() => {
+    let error;
+    for (let i = 0; i < 10; i++) {
+      i === 0
+        ? ConduitGrpcSdk.Logger.log('Attempting to connect to database...')
```
kon14, line 262: We probably shouldn't be that verbose about this. Just log the attempt once, then log failure if all attempts fail.

modules/database/src/adapters/sequelize-adapter/index.ts
```diff
@@ -255,15 +256,24 @@ export class SequelizeAdapter extends DatabaseAdapter<SequelizeSchema> {
...
         ConduitGrpcSdk.Logger.log('Sequelize connection established successfully');
         return;
-      })
-      .catch(err => {
-        ConduitGrpcSdk.Logger.error('Unable to connect to the database: ', err);
-        throw err;
-      });
+      } catch (err) {
+        error = err;
+        await sleep(2000);
```
kon14, line 270: That's a huge timeout. In cases were a db engine container is first run slightly after bringing up `Database` we're probably going to end up needlessly delaying Conduit's initialization by 2-4 secs over the db engine barely not making it in time.

kon14 (changes requested):

kon14:
Actually, let's also make the timeout duration consistent between Mongoose and Sequelize and display a connection attempt message in Mongoose.

kon14 (approved):

kkopanidis (changes requested):

modules/database/src/adapters/DatabaseAdapter.ts
```diff
@@ -10,10 +10,16 @@ type _ConduitSchema = Omit<ConduitSchema, 'schemaOptions'> & {
   extensions: DeclaredSchemaExtension[];
 };
 export abstract class DatabaseAdapter<T extends Schema> {
+  protected readonly maxConnTimeoutMs: number;
   registeredSchemas: Map<string, ConduitSchema>;
   models: { [name: string]: T } = {};
   foreignSchemaCollections: Set<string> = new Set([]); // not in DeclaredSchemas
 
+  protected constructor() {
+    this.maxConnTimeoutMs = parseInt(process.env.MAX_CONN_TIMEOUT_MS ?? '20000');
```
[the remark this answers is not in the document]
kon14: We should also default back to 5 on negative numbers just in case.

modules/database/src/adapters/sequelize-adapter/index.ts
```diff
@@ -255,15 +256,21 @@ export class SequelizeAdapter extends DatabaseAdapter<SequelizeSchema> {
   }
 
   async ensureConnected() {
-    return this.sequelize
-      .authenticate()
-      .then(() => {
+    let error;
```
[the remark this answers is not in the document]
kon14: We could break for `28P01` (passwd auth failure) or alternatively only loop while returned code matches `ECONNREFUSED`.
I'm leaning towards the latter.

kon14:
This needs to be rebased against Mongoose updates from `main`.

PR #233, opened 2022-07-13
Database admin routes register without router

ChrisPdgn:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/database/src/Database.ts
+++ b/modules/database/src/Database.ts
@@ -140,6 +140,13 @@ export default class DatabaseModule extends ManagedModule<void> {
         this._activeAdapter,
         null,
       );
+      this.adminRouter = new AdminHandlers(
+        this.grpcServer,
+        this.grpcSdk,
+        this._activeAdapter,
+        schemaController,
+        customEndpointController,
+      );
       this.grpcSdk
         .waitForExistence('router')
         .then(() => {
@@ -150,13 +157,6 @@ export default class DatabaseModule extends ManagedModule<void> {
           );
           schemaController.setRouter(this.userRouter);
           customEndpointController.setRouter(this.userRouter);
-          this.adminRouter = new AdminHandlers(
-            this.grpcServer,
-            this.grpcSdk,
-            this._activeAdapter,
-            schemaController,
-            customEndpointController,
-          );
         })
         .catch(e => {
           ConduitGrpcSdk.Logger.error(e.message);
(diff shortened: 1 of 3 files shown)


kkopanidis (changes requested):

modules/database/src/Database.ts
```diff
@@ -296,7 +298,7 @@ export default class DatabaseModule extends ManagedModule<void> {
   async setSchemaExtension(call: CreateSchemaExtensionRequest, callback: SchemaResponse) {
     try {
       const schemaName = call.request.extension.name;
-      const extOwner = (call.metadata!.get('module-name')![0] as string) as string;
+      const extOwner = call.metadata!.get('module-name')![0] as string as string;
```
kon14, line 301: Not sure why that double cast was previously there in the first place, but a single one should suffice.

modules/database/src/controllers/customEndpoints/customEndpoint.controller.ts
```diff
@@ -44,8 +50,11 @@ export class CustomEndpointController {
           CustomEndpointHandler.addNewCustomOperationControl(schema);
         });
 
-        this.router.addRoutes(routes, false);
-        this.router.requestRefresh();
+        this.router!.addRoutes(routes, false);
+        this.router!.requestRefresh();
+      })
+      .then((r: any) => {
```
kon14, line 56: Use `ICustomEndpoint` for this type.

modules/database/src/controllers/cms/schema.controller.ts
```diff
@@ -30,7 +31,15 @@ export class SchemaController {
     });
   }
 
+  setRouter(router: DatabaseRoutes) {
+    this.router = router;
+    this.refreshRoutes();
+  }
+
   refreshRoutes() {
+    if (this.router === null) {
```
kon14, line 40: We probably shouldn't warn for these. Just return from the function.
Same for rest of these checks.

kon14 (changes requested):

modules/database/src/controllers/customEndpoints/customEndpoint.controller.ts
kon14: My bad, we don't actually need the response at all.

kon14 (changes requested):

kkopanidis (approved):

kon14 (approved):

PR #259, opened 2022-07-22
fix(hermes): Swagger route tags

ChrisPdgn:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (chore)

- [ ] Yes
- [x] No


libraries/hermes/src/Rest/Swagger.ts
```diff
@@ -42,9 +42,13 @@ export class SwaggerGenerator {
   addRouteSwaggerDocumentation(route: ConduitRoute) {
     const method = this._extractMethod(route.input.action);
     let serviceName = route.input.path.toString().replace('/hook', '').slice(1);
-    serviceName = serviceName.substr(0, serviceName.indexOf('/'));
-    if (serviceName.trim() === '') {
+    let prefix = serviceName.substring(0, serviceName.indexOf('/'));
```
kon14, line 45: Make this a constant.

kon14 (changes requested):

libraries/hermes/src/Rest/Swagger.ts
kon14: Actually no, the previous line is what prefix should be initialized as.
Service name should be declared early, but assigned an initial value in the if / else blocks below.

kon14 (approved):

kkopanidis (approved):

PR #264, opened 2022-07-25
feat(authentication): resend verification email 

ChrisPdgn:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**
Resend verification email when creating a new user

### DIFF
--- a/modules/authentication/src/handlers/local.ts
+++ b/modules/authentication/src/handlers/local.ts
@@ -786,7 +786,7 @@ export class LocalHandlers implements IAuthenticationStrategy {
     });
     if (isNil(user)) throw new GrpcError(status.NOT_FOUND, 'User not found');
     if (user.isVerified)
-      throw new GrpcError(status.FAILED_PRECONDITION, 'User already exists');
+      throw new GrpcError(status.FAILED_PRECONDITION, 'User already verified');
 
     let verificationToken: Token | null = await Token.getInstance().findOne({
       type: TokenType.VERIFICATION_TOKEN,
@@ -798,10 +798,11 @@ export class LocalHandlers implements IAuthenticationStrategy {
       );
       if (diffInMilliSec < 600000) {
         const remainTime = Math.ceil((600000 - diffInMilliSec) / 60000);
-        return (
+        throw new GrpcError(
+          status.RESOURCE_EXHAUSTED,
           'Verification code not sent. You have to wait ' +
-          remainTime +
-          ' minutes to try again'
+            remainTime +
+            ' minutes to try again',
         );
       }
       await Token.getInstance().deleteMany({
--- a/modules/authentication/src/handlers/local.ts
+++ b/modules/authentication/src/handlers/local.ts
@@ -161,18 +161,20 @@ export class LocalHandlers implements IAuthenticationStrategy {
       this.verifyChangeEmail.bind(this),
     );
 
-    routingManager.route(
-      {
-        path: '/local/resend-verification',
-        action: ConduitRouteActions.POST,
-        description: `Used to resend email verification after new user is created`,
-        bodyParams: {
-          email: ConduitString.Required,
+    if (this.sendVerificationEmail) {
+      routingManager.route(
+        {
+          path: '/local/resend-verification',
+          action: ConduitRouteActions.POST,
+          description: `Used to resend email verification after new user is created`,
+          bodyParams: {
+            email: ConduitString.Required,
+          },
         },
-      },
-      new ConduitRouteReturnDefinition('ResendVerificationEmailResponse', 'String'),
-      this.resendVerificationEmail.bind(this),
-    );
+        new ConduitRouteReturnDefinition('ResendVerificationEmailResponse', 'String'),
+        this.resendVerificationEmail.bind(this),
+      );
+    }
 
     if (config.twofa.enabled) {
       routingManager.route(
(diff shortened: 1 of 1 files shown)


modules/authentication/src/handlers/local.ts
```diff
@@ -161,6 +161,19 @@ export class LocalHandlers implements IAuthenticationStrategy {
       this.verifyChangeEmail.bind(this),
     );
 
+    routingManager.route(
+      {
+        path: '/local/resend-verification',
```
kon14, line 166: Make this `/local/new/send-verification` as it's not clear what sort of verification send this URI performs.
Update handler name to reflect this.

This should also only get registered whenever `this.sendVerificationEmail` is true.

modules/authentication/src/handlers/local.ts
```diff
@@ -756,6 +773,45 @@ export class LocalHandlers implements IAuthenticationStrategy {
     return 'Email changed successfully';
   }
 
+  async resendVerificationEmail(
+    call: ParsedRouterRequest,
+  ): Promise<UnparsedRouterResponse> {
+    if (!this.sendVerificationEmail) return 'Configuration not set to send email';
```
kon14, line 779: We shouldn't be registering this route in that case.

modules/authentication/src/handlers/local.ts
```diff
@@ -756,6 +773,45 @@ export class LocalHandlers implements IAuthenticationStrategy {
...
+    let { email } = call.request.params;
+    if (AuthUtils.invalidEmailAddress(email)) {
+      throw new GrpcError(status.INVALID_ARGUMENT, 'Invalid email address provided');
+    }
+    email = email.toLowerCase();
+    const user: User | null = await User.getInstance().findOne({
+      email: email,
+    });
+    if (isNil(user)) throw new GrpcError(status.NOT_FOUND, 'User not found');
+    const verificationToken: Token | null = await Token.getInstance().findOne({
```
kon14, line 790: We should probably recreate these instead, replacing any existing ones of that type for target user.
What if we let these expire down the line?
What if token creation failed for whatever reason, in which case we're stuck with a non-verifiable user.

modules/authentication/src/handlers/local.ts
```diff
@@ -756,6 +773,45 @@ export class LocalHandlers implements IAuthenticationStrategy {
...
+    });
+    if (isNil(verificationToken)) {
+      throw new GrpcError(status.NOT_FOUND, 'Verification email token does not exist');
+    }
+    const serverConfig = await this.grpcSdk.config.get('router');
+    const url = serverConfig.hostUrl;
+    const result = { verificationToken, hostUrl: url };
+    const link = `${result.hostUrl}/hook/authentication/verify-email/${result.verificationToken.token}`;
+    await this.emailModule
+      .sendEmail('ChangeEmailVerification', {
```
kkopanidis, line 802: you're using the change email verification instead of the email verification template

kkopanidis (changes requested):

modules/authentication/src/handlers/local.ts
```diff
@@ -756,6 +773,45 @@ export class LocalHandlers implements IAuthenticationStrategy {
...
+        email: email,
+        sender: 'no-reply',
+        variables: {
+          link,
+        },
+      })
+      .catch(e => {
+        ConduitGrpcSdk.Logger.error(e);
+      });
+    return 'Verification email resent';
```
kon14, line 812: Make this `Verification code sent`.
We shouldn't treat this as a resend route/function.

kon14 (changes requested):

modules/authentication/src/handlers/local.ts
```diff
@@ -161,18 +161,20 @@ export class LocalHandlers implements IAuthenticationStrategy {
...
     );
 
-    routingManager.route(
-      {
-        path: '/local/resend-verification',
-        action: ConduitRouteActions.POST,
-        description: `Used to resend email verification after new user is created`,
-        bodyParams: {
-          email: ConduitString.Required,
+    if (this.sendVerificationEmail) {
```
kon14, line 164: This should only register whenever `this.sendVerificationEmail` is true.
Move this into the relevant block above.

modules/authentication/src/handlers/local.ts
```diff
@@ -756,6 +775,60 @@ export class LocalHandlers implements IAuthenticationStrategy {
...
+    let { email } = call.request.params;
+    if (AuthUtils.invalidEmailAddress(email)) {
+      throw new GrpcError(status.INVALID_ARGUMENT, 'Invalid email address provided');
+    }
+    email = email.toLowerCase();
+    const user: User | null = await User.getInstance().findOne({
+      email: email,
+    });
+    if (isNil(user)) throw new GrpcError(status.NOT_FOUND, 'User not found');
+    if (user.isVerified) return 'User is already verified';
```
kon14, line 790: This should throw an error instead.

kon14 (changes requested):

modules/authentication/src/handlers/local.ts
```diff
@@ -787,21 +785,13 @@ export class LocalHandlers implements IAuthenticationStrategy {
       email: email,
     });
     if (isNil(user)) throw new GrpcError(status.NOT_FOUND, 'User not found');
-    if (user.isVerified) return 'User is already verified';
+    if (user.isVerified)
+      throw new GrpcError(status.FAILED_PRECONDITION, 'User already exists');
```
kon14, line 789: `User is already verified` was fine. I only requested that the return be changed to a throw.

kon14 (changes requested):

kon14 (approved):

kkopanidis (approved):

PR #267, opened 2022-07-26
fix(admin,commons,core,grpc-sdk,router,storage): config updates

kon14:
Fixes `Admin` not handling live config updates.
Fixes `Router` and `Storage` using last configuration during `onConfig()` calls.

We could also fix the Convict schema type, but doing so in `Commons` would require importing the config schema from `Admin`.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No

### DIFF
--- a/libraries/grpc-sdk/src/modules/config/index.ts
+++ b/libraries/grpc-sdk/src/modules/config/index.ts
@@ -48,17 +48,6 @@ export class Config extends ConduitModule<typeof ConfigDefinition> {
     });
   }
 
-  updateConfig(config: any, name: string) {
-    const request = {
-      config: JSON.stringify(config),
-      moduleName: name,
-    };
-
-    return this.client!.updateConfig(request).then(res => {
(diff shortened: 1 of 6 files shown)


Michael-Vol (approved):

libraries/grpc-sdk/src/modules/config/index.ts
```diff
@@ -48,17 +48,6 @@ export class Config extends ConduitModule<typeof ConfigDefinition> {
     });
   }
 
-  updateConfig(config: any, name: string) {
```
kkopanidis, line 51: This function continues to exist in core.proto. Either remove it from there to (and from config manager) or don't delete it from the sdk

kkopanidis (changes requested):

kkopanidis (approved):

PR #272, opened 2022-07-28
refactor(authentication): email verification link rework

ChrisPdgn:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/authentication/src/handlers/local.ts
+++ b/modules/authentication/src/handlers/local.ts
@@ -716,7 +716,9 @@ export class LocalHandlers implements IAuthenticationStrategy {
     });
 
     if (isNil(verificationTokenDoc)) {
-      const redisToken = await this.grpcSdk.state!.getKey(verificationTokenParam);
+      const redisToken = await this.grpcSdk.state!.getKey(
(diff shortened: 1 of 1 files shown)


modules/authentication/src/handlers/local.ts
```diff
@@ -734,13 +738,18 @@ export class LocalHandlers implements IAuthenticationStrategy {
       user,
     );
     const tokenPromise = Token.getInstance().deleteOne(verificationTokenDoc);
+    await this.grpcSdk.state!.setKey(
```
kon14, line 741: I don't like how these tokens are just plainly dumped straight into Redis, but I guess there's no other way around it as Redis does not support expiring individual hash keys.

Can we at the very least prefix these with an additional `_verifiedToken` or something?
That would result in Redis storing these as `authentication_verifiedToken_XYZXYZXYZ`.

kon14 (changes requested):

kkopanidis (approved):

PR #281, opened 2022-08-10
fix(admin): auth middleware body query check

kon14:
Fixes query param check failing for REST requests containing a `query` body param.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No

### DIFF
--- a/packages/admin/src/middleware/Auth.middleware.ts
+++ b/packages/admin/src/middleware/Auth.middleware.ts
@@ -26,7 +26,7 @@ async function requestExcluded(req: ConduitRequest, conduit: ConduitCommons) {
   if (excludedRestRoutes.includes(req.path)) return true;
   // GraphQL
   if (
-    req.originalUrl.startsWith('/graphql/') &&
+    req.originalUrl.startsWith('/graphql') &&
     req.body.query &&
     typeof req.body.query === 'string'
   ) {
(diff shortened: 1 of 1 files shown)


Michael-Vol (approved):

packages/admin/src/middleware/Auth.middleware.ts
```diff
@@ -20,8 +20,14 @@ async function requestExcluded(req: ConduitRequest, conduit: ConduitCommons) {
   if (excludedRoutes.includes(req.path)) return true;
   // GraphQL
   if (excludedGqlMutations.includes(req.body.operationName)) return true;
-  const queryName = req.body.query?.split('{\n')[1].split('\n}')[0].trim();
-  if (excludedGqlQueries.includes(queryName)) return true;
+  if (
+    req.body.query &&
+    typeof req.body.query === 'string' &&
+    req.body.query.startsWith('{\n')
```
kkopanidis, line 26: These checks require the existence of a special character "\n" to work. This doesn't seem stable at all, we don't care about general content just the existence of a specific query. This check needs to be changed to reflect that in a more concrete manner

kkopanidis (changes requested):

kkopanidis (approved):

PR #286, opened 2022-08-24
feat(hermes,grpc-sdk,database): db model router type registration

kon14:
This PR facilitates the usage of db model types as router return types by automatically registering a model's fields in a type registry to be retrieved whenever coming across an unknown route return definition name.

~This **does not** take into account default model field selectors, meaning an automatically imported `User` type would show up in route docs with a `hashedPassword` field being supposedly included, despite the db adapter not returning it.~

This takes into account model field selectors, meaning unselected fields (eg `User.hashedPassword`) don't show up in route types.

**Explicitly declared types**, defined as return types for other registered routes **take precedence over implicit model types**.
It's also important to remember how **administrative and client application APIs do not share explicit route return type declarations.**

Example:
`AccessToken` is not currently declared as an administrative route return type name.
If I were to reference it as an administrative route's return type, Hermes would attempt to pull it from the db type registry.
The `Authentication` module registers `AccessToken` as one of its db models, hence we'd succeed in retrieving its type and returning that as the route's return type.
Now, if I were to explicitly register an administrative route with a return type name of `AccessToken` where I'd specify my own fields, that one would take precedence over the db type and be used as the formerly discussed route's return definition as well.

This also introduces support for _implicit_ (for lack of a better name) route return type declarations.
Example: A route returning a single chat message (db schema name: `ChatMessage`) typically features a return type definition like this: `new ConduitRouteReturnDefinition('ChatMessage', ChatMessage.getInstance().fields)`

This may now simply be written as `new ConduitRouteReturnDefinition('ChatMessage')`
or even `new ConduitRouteReturnDefinition('RenamedChatMessageType', 'ChatMessage')`

**Edit:**
I rebased this PR against `main` and made some big changes to the type registry functionality.
Types are now pulled directly from the database (`DeclaredSchema`) instead of going through `Redis`.
There's still a couple of things left to do:
- [X] Remove `Core`'s startup dependency on `Database` reintroduced for initial type lookup.
- [X] Proc a doc type reparsing across enabled `Admin`/`Router` transports on schema extension modifications.
- [X] ~Update `Database`'s gRPC functions to expose `compiledFields` to be used in `TypeRegistry` (probably in a followup PR)~ Expose compiled schema fields (base + extensions) (#313).

**Notes:**
Upon altering a module's route return definitions (**in code**), **`Core` and `Router` should be restarted** in order for the changes to take effect in the administrative and application level APIs.
Redis state may prevent changes from taking place.
This is not a regression introduced in this PR and should mostly be considered while debugging as both `Core` and `Router` should generally be restarted anyway while updating to new Conduit release.
I'm saying mostly, as it's most likely going to affect anyone attempting to reuse generated route return types.

**Limitations & Optimizations:**
- `TypeRegistry` has no way of removing `requestedTypes` entries once the routes requesting them are removed (or altered so that they don't need these types anymore), resulting in the transport update handlers getting executed whenever these schema types get updated in the database.
We should ideally be removing these in a future iteration of this.

GraphQl aborts its server refresh on type parsing failure as schemas are enforced.

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No


Michael-Vol (approved):

ChrisPdgn (approved):

kon14:
I updated this PR, rebasing it against the latest `main` and modifying it so as to pull types directly from the `Database` instead of registering them in Redis first.

Check the OP for remaining TODOs and whatnot.

libraries/hermes/src/classes/TypeRegistry.ts
```diff
@@ -0,0 +1,97 @@
...
+        });
+    });
+  }
+
+  async setType(schemaName: string) {
+    if (!this.dbAvailable) {
+      this.pendingTypes.add(schemaName);
+      return;
+    }
+    await sleep(250); // prevent gRPC request flooding during early Core/Database initialization
```
kon14, line 54: This should be addressed before merge.

libraries/grpc-sdk/src/classes/ConduitRouteReturn.ts
```diff
@@ -42,11 +42,23 @@ import { ConduitModel } from '../interfaces';
...
+  // - registered database schema types
+
+  constructor(
+    name: string,
+    fields?: ConduitModel | string, //
+  ) {
     this._name = name;
-    this._fields = fields;
+    // TODO:
+    // Defaulting to named schema types should throw on "reserved" base types (throw on Hermes-side also).
```
kon14, line 58: This already affects `main`, but it's out of scope for this PR.

kon14:
I'm investigating a weird interaction with `Storage` where bringing it up causes `Admin` to fail to retrieve the following config-related non-types:
- `local`
- `conduit`
- `(empty string)`
- `/var/tmp`

kkopanidis (changes requested):

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -265,6 +265,7 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
 
     delete this.models[schemaName];
     this.mongoose.connection.deleteModel(schemaName);
+    this.grpcSdk.bus!.publish('database:dataTypes:deregistration', schemaName);
```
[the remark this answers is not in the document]
kon14: What if we don't delete the schema from gRPC? eg CMS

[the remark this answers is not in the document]
kkopanidis: ok then, the first point still stands, the same "if we don't do it from CMS" is valid for create schema too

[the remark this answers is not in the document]
kon14: We should probably reuse a single event type consistently emitted from the adapters regardless of this PR 👍.

libraries/hermes/src/classes/TypeRegistry.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export class TypeRegistry {
+  private static instance: TypeRegistry;
+  private static transportUpdateHandlers: Map<string, TransportUpdateHandler> = new Map();
+  private readonly schemaTypes: Map<string, ConduitModel> = new Map();
+  private readonly requestedTypes: Set<string> = new Set();
+  private dbAvailable = false;
+
+  private constructor(private readonly grpcSdk: ConduitGrpcSdk) {
+    this.waitForDatabase().then();
```
ChrisPdgn, line 18: Empty then()

kon14: That is intentional, we already handle resolution internally. It's just explicit.

ChrisPdgn (approved):

kkopanidis (approved):

PR #299, opened 2022-08-31
refactor: rename schemaOptions, modelSchema to modelOptions, fields

kon14:
This PR simplifies schema types, renaming certain fields for consistency and readability.
Notably, `modelOptions` and `schemaOptions` are finally using the same field name 🎉.
Ideally, we'd simply use `options`, but we're sticking with `modelOptions` (for now?) due to a complication with Sequelize migrations.

`modelSchema` has also been renamed to `fields` for readability.

This is technically a Breaking Change as `Database`'s gRPC definitions have been slightly altered because of the above.
Routing APIs shouldn't be affected by this.

This **does not resolve all schema related type issues**.
We still need to work on type anys and invalid type usages (eg `ConduitSchema`-like types with stringified fields being advertised as `ConduitSchema`) etc.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [X] Yes
- [ ] No

### DIFF
--- a/modules/database/src/adapters/mongoose-adapter/index.ts
+++ b/modules/database/src/adapters/mongoose-adapter/index.ts
@@ -18,7 +18,7 @@ import { mongoSchemaConverter } from '../../introspection/mongoose/utils';
 const parseSchema = require('mongodb-schema');
 let deepPopulate = require('mongoose-deep-populate');
 
-type _ConduitSchema = Omit<ConduitSchema, 'schemaOptions'> & {
+type _ConduitSchema = Omit<ConduitSchema, 'modelOptions'> & {
   modelOptions: ConduitSchemaOptions;
 };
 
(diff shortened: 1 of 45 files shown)


kkopanidis (approved):

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -18,8 +18,8 @@ import { mongoSchemaConverter } from '../../introspection/mongoose/utils';
 const parseSchema = require('mongodb-schema');
 let deepPopulate = require('mongoose-deep-populate');
 
-type _ConduitSchema = Omit<ConduitSchema, 'schemaOptions'> & {
-  modelOptions: ConduitModelOptions;
+type _ConduitSchema = Omit<ConduitSchema, 'modelOptions'> & {
```
ChrisPdgn, line 21: Remove unused type

ChrisPdgn (changes requested):

PR #317, opened 2022-09-08
Core rpc service tests

sdimitris:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -7,6 +7,9 @@
   "engines": {
     "node": ">=14"
   },
+  "jest": {
+    "testTimeout": 20000
+  },
   "scripts": {
     "start": "node dist/bin/www.js",
     "lint": "./node_modules/.bin/eslint .",
@@ -15,7 +18,8 @@
     "build": "rimraf dist && tsc",
     "postbuild": "copyfiles -u 1 src/*.proto ./dist/",
     "prepare": "npm run build",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build",
+    "tests": "jest"
   },
   "license": "ISC",
   "dependencies": {
--- a/packages/core/tests/grpc/config.test.js
+++ b/packages/core/tests/grpc/config.test.js
@@ -1,13 +1,13 @@
 let client;
+let core;
 describe('Testing Config package', () => {
-  beforeAll(() => {
+  beforeAll(async () => {
     const { exec } = require('child_process');
     const options = {
       env: { ...process.env, REDIS_PORT: 6379, REDIS_HOST: 'localhost', PORT: 3030, ADMIN_SOCKET_PORT: 3032 }
     }
-    exec("node ./dist/bin/www.js", options,function (err,stdout,stderr) {
-
-    });
+    core = exec("node ./dist/bin/www.js", options);
+    await new Promise((r) => setTimeout(r, 2000));
     const path = require('path');
     const protoLoader = require('@grpc/proto-loader');
     const grpc = require('@grpc/grpc-js');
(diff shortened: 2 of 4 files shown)


packages/core/src/tests/grpc/config.test.js
```diff
@@ -1,20 +1,26 @@
-let client;
-let coreProcess;
-
+let coreProcess,client,testModule;
 beforeAll(async () => {
   const { exec } = require('child_process');
-  const options = {
+  let options = {
     env: { ...process.env, REDIS_PORT: 6379, REDIS_HOST: 'localhost', PORT: 3030, ADMIN_SOCKET_PORT: 3032 },
```
kon14, line 5: Reorder these so that process env values override the defaults whenever available.

packages/core/src/tests/grpc/config.test.js
```diff
@@ -1,20 +1,26 @@
...
+  exec('sh ./src/tests/scripts/redis.sh');
   await new Promise((r) => setTimeout(r, 2000));
+  exec('cp ./src/tests/mocks/test.proto ./dist/tests/mocks/module')
+  await new Promise((r) => setTimeout(r, 3000))
   coreProcess = exec('node ./dist/bin/www.js', options);
   await new Promise((r) => setTimeout(r, 5000));
-
+  options = {
+    env: {
+      CONDUIT_SERVER: '0.0.0.0:55152', SERVICE_IP: '0.0.0.0:55184'
```
kon14, line 15: These should accept env values whenever available, just like above.

kon14 (changes requested):
Eslint is not parsing this somehow. Run manually.

libraries/testing-tools/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    const channel = createChannel(this.serverAddress, undefined, {
+      'grpc.max _receive_message_length': 1024 * 1024 * 100,
+      'grpc.max_send_message_length': 1024 * 1024 * 100,
+    });
+    const clientFactory = createClientFactory().use(getModuleNameInterceptor('test'));
+    this._client = clientFactory.create(serviceDefinition, channel);
+  }
+
+  async startRedis() {
+    exec('docker run --name conduit-redis -d -p 6379:6379 redis:latest');
```
kon14, line 22: Add `REDIS_PORT` env for configuring the exported port and default to something else.
Otherwise running any of these tests locally while working with Conduit would require constantly shutting down our dev Redis.

This also doesn't just **start** Redis, it creates it too.
Change `conduit-redis` to `conduit-redis-tests` or something that shouldn't normally collide with anything else.
Depending on how these functions are used we may need to check whether the container already exists and start it instead as otherwise we'd be constantly recreating containers.

libraries/testing-tools/src/interfaces/IRRunDependenciesInterface.ts
```diff
@@ -0,0 +1,9 @@
+export interface IRRunDependenciesInterface {
```
kon14, line 1: What does first R in IRRun stand for?

kon14 (changes requested):

kon14 (dismissed):
Tests shouldn't be automatically built together with their corresponding packages as that will increase build times.
We just need tests to be built and run on PRs.

kkopanidis:
@ChrisPdgn generally looks good I think. 2 points we need to fix: Coverage, and wait for the full release of 0.15

kkopanidis (approved):

PR #322, opened 2022-09-12
feat: route doc descriptions

ChrisPdgn:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No


modules/database/src/admin/index.ts
```diff
@@ -305,6 +316,7 @@ export class AdminHandlers {
         {
           path: '/schemas/:schemaId/extensions',
           action: ConduitRouteActions.POST,
+          description: `Sets extension fields to a schema.`,
```
kon14, line 319: This one actually sets, updates and deletes the `Database` module's own extension** for that schema.
You could phrase this like so:
`Sets a Database-owned extension for target schema, expanding it with additional fields. Passing an empty fields object removes the extension.`

modules/database/src/admin/index.ts
```diff
@@ -351,6 +365,7 @@ export class AdminHandlers {
         {
           path: '/introspection',
           action: ConduitRouteActions.POST,
+          description: `Performs database introspection.`,
```
kon14, line 368: We should ideally be mentioning pending schemas here as we reference them in followup descriptions.
I would suggest something along the lines of:
`Performs database introspection, registering any unknown collections as pending schemas.`

modules/database/src/admin/index.ts
```diff
@@ -386,6 +403,7 @@ export class AdminHandlers {
         {
           path: '/introspection/schemas/finalize',
           action: ConduitRouteActions.POST,
+          description: `Finalizes schema structures upon introspection.`,
```
kon14, line 406: `Converts a previously imported pending schema to a CMS schema.`

modules/email/src/admin/index.ts
```diff
@@ -83,6 +84,7 @@ export class AdminHandlers {
         {
           path: '/templates',
           action: ConduitRouteActions.POST,
+          description: `Creates a new email template requiring at least template's name and body.`,
```
kon14, line 87: Shouldn't elaborate on param requirements.
This can change independently of the description and are already marked as required or optional in the route doc.

modules/email/src/admin/index.ts
```diff
@@ -101,6 +103,7 @@ export class AdminHandlers {
         {
           path: '/templates/:id',
           action: ConduitRouteActions.PATCH,
+          description: `Updates specified fields of a template and requires its id.`,
```
kon14, line 106: and part is redundant

packages/admin/src/routes/ChangePasswordRoute.ts
```diff
@@ -15,6 +15,7 @@ export function changePasswordRoute() {
     {
       path: '/change-password',
       action: ConduitRouteActions.POST,
+      description: `Changes admin user's password.`,
```
kon14, line 18: `Changes authenticated admin user's password`

packages/core/src/config-manager/admin/routes/GetConfigRoute.ts
```diff
@@ -13,6 +13,7 @@ export default function getConfigRoute(
     {
       path: `/config/${moduleName}`,
       action: ConduitRouteActions.GET,
+      description: `Returns configuration of module ${moduleName}.`,
```
kon14, line 16: `Returns configuration of ${moduleName} module.`

packages/core/src/config-manager/admin/routes/GetModules.route.ts
```diff
@@ -11,6 +11,7 @@ export function getModulesRoute(registeredModules: Map<string, RegisteredModule>
     {
       path: '/config/modules',
       action: ConduitRouteActions.GET,
+      description: `Returns configuration of all registered modules.`,
```
kon14, line 14: This actually returns an array of module objects, containing their name, url and serving status.
You could phrase this like:
`Returns as array of currently available modules, providing information about their service`

modules/authentication/src/handlers/local.ts
```diff
@@ -81,6 +81,7 @@ export class LocalHandlers implements IAuthenticationStrategy {
         {
           path: '/forgot-password',
           action: ConduitRouteActions.POST,
+          description: `Used when user forgets password. Sends an email for password reset.`,
```
kon14, line 84: That would more accurately be something like: `Resets the authenticated user's password. Sends a verification email if email verification is enabled AND verification emails are enabled AND the Email module available.`
That's a mouthful and still doesn't actually explain everything there is to be explained.
We respond with an appropriate message about verification being required, therefore we should probably keep it simple:

`Resets the authenticated user's password.`

To be honest we shouldn't be mentioning conditional config options at all unless enabled, just like for security clients, but propagating that info down here would be kinda meh unless done in a seemless way.

kon14 (dismissed):
Remove explicit references to required params in descriptions (eg `requiring only ...`) as these can become outdated and are otherwise already documented for each route outside the description.

Remove the `all` part from getters (eg `Returns all forms`) as most of them can be queried, or at the very least paginated, based on other params etc.

modules/authentication/src/handlers/local.ts
ChrisPdgn: Either way, this endpoint is exposed only when sendVerificationEmail config is enabled, so we could be a little more specific in the description I guess?

kon14: Fair enough `Generates a password reset token and forwards a verification link to the user's email address.`

kkopanidis (approved):

PR #347, opened 2022-09-23
feat: add node-2fa library

ChrisPdgn:
The creation of this new library is meant to help with occurring problems and reduce external dependencies.
Subsequently, this library will be used from an upcoming central 2fa handler, in order to simplify 2fa process. 
So any duplicate code and other redundant stuff will be dealt with soon.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No


modules/authentication/src/handlers/local.ts
```diff
@@ -719,6 +719,10 @@ export class LocalHandlers implements IAuthenticationStrategy {
       const verification = TwoFactorAuth.verifyToken(secret.secret, code, 1);
       if (isNil(verification))
         throw new GrpcError(status.UNAUTHENTICATED, 'Verification unsuccessful');
+      if (verification.delta === -1)
```
kkopanidis, line 722: do not provide extra codes. If delta is anything else besides 0, provide a 400 error (wrong params) and a message that the code is not correct

modules/authentication/src/handlers/local.ts
```diff
@@ -979,6 +983,10 @@ export class LocalHandlers implements IAuthenticationStrategy {
     if (isNil(verification)) {
       throw new GrpcError(status.UNAUTHENTICATED, 'Verification unsuccessful');
     }
+    if (verification.delta === -1)
```
kkopanidis, line 986: do not provide extra codes. If delta is anything else besides 0, provide a 400 error (wrong params) and a message that the code is not correct

libraries/node-2fa/package.json
```diff
@@ -0,0 +1,33 @@
+{
```
kkopanidis, line 1: Add here the author details from the original repo. We need to make sure we give credit to the author

kkopanidis (changes requested):

kon14 (approved):

PR #404, opened 2022-10-24
feat(authentication): 2fa backup codes

ChrisPdgn:
This PR introduces the generation of back up codes for 2FA.
When authenticated, the user can create 10 one-time-use codes for access recovering, when 2FA isn't possible at that moment.
After being used, each code gets deleted, so the user has to generate more when only a few of them are left. 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/authentication/src/handlers/twoFa.ts
+++ b/modules/authentication/src/handlers/twoFa.ts
@@ -137,8 +137,7 @@ export class TwoFa implements IAuthenticationStrategy {
         {
           path: '/twoFa/recover',
           action: ConduitRouteActions.POST,
-          description: `Recover 2FA access with a back up code. 
-                      The format of one code is '4 digits (space) 4 digits'.`,
+          description: `Recovers 2FA access with an 8 digit backup code.`,
           middlewares: ['authMiddleware'],
           bodyParams: {
             code: ConduitString.Required,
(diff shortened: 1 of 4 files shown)


modules/authentication/src/models/TwoFactorBackUpCodes.schema.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+  },
+} as const;
+const collectionName = undefined;
+
+export class TwoFactorBackUpCodes extends ConduitActiveSchema<TwoFactorBackUpCodes> {
+  private static _instance: TwoFactorBackUpCodes;
+  _id: string;
+  user: string | User;
+  codes: string[];
```
kon14, line 33: Why generate multiple codes in the first place?
Is that merely so we can invalidate used ones upon recovery without worrying about the user not receiving/recreating new ones?

Regardless, why not use a 1:N relation instead of storing these in a single array?

kon14 (changes requested):

kon14 (approved):

modules/authentication/src/handlers/twoFa.ts
```diff
@@ -364,6 +399,65 @@ export class TwoFa implements IAuthenticationStrategy {
...
+      throw new GrpcError(status.UNAUTHENTICATED, 'Unauthorized');
+    }
+    return this.codeGenerator(context.user);
+  }
+
+  async recoverTwoFa(call: ParsedRouterRequest): Promise<UnparsedRouterResponse> {
+    const context = call.request.context;
+    const { user, clientId } = context;
+    const { code } = call.request.params;
+    if (isNil(context) || isNil(user)) {
```
kkopanidis, line 420: remove context and user checks

kkopanidis (changes requested):

kkopanidis (approved):

PR #424, opened 2022-11-16
Sorting option in core,email and router routes.

sdimitris:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

Routes which support sorting option:
- /config/modules
- /router/routes
- /router/middlewares
- /email/externalTemplates

### DIFF
--- a/modules/router/src/admin/router.ts
+++ b/modules/router/src/admin/router.ts
@@ -1,11 +1,9 @@
 import ConduitGrpcSdk, {
-  GrpcError,
   ParsedRouterRequest,
   UnparsedRouterResponse,
 } from '@conduitplatform/grpc-sdk';
 import { isNil } from 'lodash';
 import ConduitDefaultRouter from '../Router';
-import { status } from '@grpc/grpc-js';
 
 export class RouterAdmin {
   constructor(
--- a/modules/router/src/admin/router.ts
+++ b/modules/router/src/admin/router.ts
@@ -14,9 +14,7 @@ export class RouterAdmin {
   ) {}
 
   async getMiddlewares(call: ParsedRouterRequest): Promise<UnparsedRouterResponse> {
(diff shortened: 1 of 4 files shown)


kkopanidis (changes requested):
if only sort by name is available, name it such. The param should be sortByName and the value should be true or false or something simple in general.

modules/router/src/admin/router.ts
```diff
@@ -1,15 +1,21 @@
-import ConduitGrpcSdk, { UnparsedRouterResponse } from '@conduitplatform/grpc-sdk';
+import ConduitGrpcSdk, {
+  GrpcError,
```
ChrisPdgn, line 2: Remove unused import

modules/router/src/admin/router.ts
```diff
@@ -1,15 +1,21 @@
...
+  ParsedRouterRequest,
+  UnparsedRouterResponse,
+} from '@conduitplatform/grpc-sdk';
 import { isNil } from 'lodash';
 import ConduitDefaultRouter from '../Router';
+import { status } from '@grpc/grpc-js';
```
ChrisPdgn, line 8: and here

packages/core/src/config-manager/admin/routes/GetModules.route.ts
```diff
@@ -4,15 +4,22 @@ import {
   ConduitError,
   ConduitString,
   ConduitBoolean,
+  GrpcError,
```
ChrisPdgn, line 7: and here

ChrisPdgn (approved):

kkopanidis (approved):

PR #442, opened 2022-11-22
fix(database): body params getting in GET requests

sdimitris:
Custom GET endpoints can not have body parameters.
Before this fix , body parameters could be given to a custom GET endpoint.
So, this fix checks if body parameters have been given to a custom GET request.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/database/src/admin/customEndpoints/utils.ts
+++ b/modules/database/src/admin/customEndpoints/utils.ts
@@ -156,11 +156,14 @@ function _inputValidation(
   }
 
   if (location === 2 && isArray) {
-    return 'Url params cant have an array input';
+    return "Url params can't have an array input";
   }
 
-  if (operation === OperationsEnum.GET && location === 0) {
-    return 'GET requests can not have body parameters';
+  if (
+    (operation === OperationsEnum.GET || operation === OperationsEnum.DELETE) &&
+    location === 0
+  ) {
+    return 'GET or DELETE requests can not have body parameters';
   }
 
   return true;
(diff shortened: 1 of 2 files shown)


modules/database/src/admin/customEndpoints/utils.ts
```diff
@@ -131,6 +131,7 @@ function _queryValidation(
  * }
  */
 function _inputValidation(
+  operation: number,
```
kon14, line 134: While you're at it, update `operation` and `location` to use enums instead of plain numbers.
We already have an `OperationsEnum` object type that should be turned into an actual enum.

kon14 (dismissed):

modules/database/src/admin/customEndpoints/utils.ts
```diff
@@ -158,17 +159,31 @@ function _inputValidation(
     return 'Url params cant have an array input';
```
kon14, line 159: Fix typo in `cant` -> `can't`

kkopanidis (changes requested):
DELETE requests can't have body params either

modules/database/src/admin/customEndpoints/utils.ts
sdimitris: I think that should be fixed in another PR

kkopanidis (approved):

PR #447, opened 2022-11-23
feat: Conduit Manifest

kon14:
This PR introduces support for Conduit module declaration manifest objects as part of `package.json`.
Said manifest objects currently provide the following information:
- module version
- module dependencies

We should probably include support for detailing:
- a module's supported features
- feature dependencies
- dependencies on features
- migration state

Modules attempting to register themselves with Conduit provide their manifest file along with the registration request.
They are initially marked as `pending`, until their dependencies are met, at which point they receive a standardized activation gRPC request (`activateModule()`), allowing them to be properly brought up etc.

Also fixes `grpc-sdk`'s `isModuleUp()`.

**BREAKING CHANGE:**

Modified Conduit Config Service RPCs:
-Dropped ModuleExists()
-Dropped ModuleList()
-Refactored WatchModules() into WatchDeploymentState()
-Updated RegisterModule()

Services need also expose an `ActivateModule()` RPC
GetModules route return type modified.

I'm opening this up for review.
**Requires further testing!**

- [X] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [X] Yes
- [ ] No


packages/core/tests/config.test.ts
```diff
@@ -53,7 +53,8 @@ describe('Testing module related rpc calls', () => {
   test('Register Module', async () => {
     const res = await testClient.registerModule({
```
kon14, line 54: I should update these.

packages/core/src/config-manager/service-discovery/index.ts
```diff
@@ -1,44 +1,70 @@
...
+  ) {
     this.moduleRegister = new EventEmitter();
     this.moduleRegister.setMaxListeners(150);
+    this.registeredModules = new Map([
+      [
+        'conduit',
+        {
+          moduleName: 'conduit',
+          moduleVersion: ManifestManager.getInstance().moduleVersion,
+          moduleUrl: 'unknown', // TODO
```
kon14, line 58: This doesn't technically break anything, but it's still an eye-sore that should probably be addressed.

modules/database/src/Database.ts
```diff
@@ -101,7 +101,7 @@ export default class DatabaseModule extends ManagedModule<void> {
     this.registerInstanceSyncEvents();
     const coreHealth = (await this.grpcSdk.core.check()) as unknown as HealthCheckStatus;
     this.onCoreHealthChange(coreHealth);
-    await this.grpcSdk.core.watch('');
+    this.grpcSdk.core.watch().then();
```
kon14, line 104: Awaiting this wouldn't work. Needs investigation.

kon14:
Updated this, moving standardized Conduit module RPCs into a separate service.
I'll update the client getters to support dynamic modules (so that eg Core can `setConfig()` out-of-tree modules).

kkopanidis (approved):

PR #471, opened 2022-12-19
feat: migration support & column deletion support

ChrisPdgn:
**This PR introduces a new strategy for executing migrations and a new feature that enables column deletion of a schema (previously was not allowed)**

**Migration execution now follows the steps below:**
- Creation of ManagedModule -> ModuleManager gets notified that a module has to perform migration by passing the absolute path of the migration file folder to manager's constructor
- The module is then registered with unknown serving state and awaits an initialization event that triggers after successful execution of all module migrations and changes its serving state accordingly

**A migration file now has the following format:**
```javascript
module.exports = {
  up: async function (grpcSdk: ConduitGrpcSdk) {
    
  },
  down: async function (grpcSdk: ConduitGrpcSdk) {
    
  }
}
```
The up function contains all migration raw queries that need to be executed when running a migration, while the down function contains all raw queries that undo the changes above.

Migratiοns can be executed automatically or manually according to the project's environment (specified in core config).
In case of manual migrations, 3 endpoints have been implemented to trigger migrations for a module, to see all manually executed migrations and to perform a downgrade on a selected migration.

Each registered migration has its own migrationLogs schema that stores a history of the migration states.

**Included fixes:**
- ManifestManager validateTag() 
- Database rawQuery() in mongoose had to convert results to array in case of find operation 
- Added missing collectionName field to getSchema(s) functions at grpc-sdk database wrapper
- Changed the sequelize sync config options to { drop: false }, a bug causing Authentication migrations to fail in sequelize. In case of column deletion another sync is performed in order to allow columns to drop

**Alterations:**

For storing the contents of a migration file, the String datatype in Sequelize had to be replaced with Text datatype. 
This needs to be addressed sooner or later.
   

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No


libraries/grpc-sdk/src/utilities/registerMigrations.ts
```diff
@@ -0,0 +1,32 @@
...
+  database: DatabaseProvider,
+  moduleName: string,
+  moduleVersion: string,
+  path: string,
+) {
+  const files = readdirSync(path).filter(f => f.endsWith('.js'));
+  const migrations = new Map<string, string>();
+  for (const name of files) {
+    // Ignore authentication's config migration for convict
+    if (name === 'configMigration.js' && moduleName === 'authentication') continue;
```
kkopanidis, line 22: instead of adding a specific file here, you should check if the file includes the up and down properties for example.

modules/authentication/src/Authentication.ts
```diff
@@ -57,9 +58,11 @@ export default class Authentication extends ManagedModule<Config> {
   async onServerStart() {
     await this.grpcSdk.waitForExistence('database');
     this.database = this.grpcSdk.database!;
-    TokenProvider.getInstance(this.grpcSdk);
     await this.registerSchemas();
-    await runMigrations(this.grpcSdk);
+    const version = ManifestManager.getInstance().moduleVersion;
```
kkopanidis, line 62: can we add the registerMigrations logic to module manager? That would be cleaner I think, and the modules can optionally specify a migrationsPath variable so that they can modify migration location.

modules/authentication/src/index.ts
```diff
@@ -4,6 +4,7 @@ import { Config } from './config';
 import path from 'path';
 
 const authentication = new AuthenticationModule();
+authentication.activateMigrations();
```
kkopanidis, line 7: what's the purpose of having activateMigrations here?

modules/authentication/src/migrations/UserIdToUserAccessTokenSchemaMigration.ts
```diff
@@ -1,17 +1,50 @@
...
+      query = {
+        sqlQuery: {
+          query: `SELECT * FROM "${sqlTableName}";`,
+        },
+      };
+      const accessTokens = await database.rawQuery('AccessToken', query);
+      if (accessTokens.length === 0) return;
+      query = {
+        sqlQuery: {
+          query: `ALTER TABLE "${sqlTableName}" ADD COLUMN IF NOT EXISTS "user" uuid;`,
```
kkopanidis, line 35: Is this required? I thought that sequelize will have already added the column(?)

kkopanidis (changes requested):

PR #490, opened 2023-01-20
feat(hermes,router,admin,core): middleware patch support

ChrisPdgn:
This PR introduces a support for adding new middleware to a route with a specific order regarding to the rest registered middleware of the route.
In order to make use of this middleware patch, first you have to register the middleware to the according router.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**


kkopanidis:
This is more of a question. @ChrisPdgn what will happen if a module for ex. Authentication refreshes its routes, and another module ex. CMS has added a middleware through patch? I'm guessing the middleware will be removed but it shouldn't.

Also there should be a way to remove a middleware and/or modify it at least through the admin panel.

PR #508, opened 2023-02-06
feat(database): like,ilike operators

kon14:
This PR adds `$like` and `$ilike` operators, introducing support for case-insensitive SQL like comparisons.
Like-strings are now wrapped in `%` instead of `.*`. Eg: `.*abc.*` -> `%abc%`.

~~Custom endpoint queries are to be migrated over to the new format.~~
Custom endpoint syntax has been updated to remain compatible with legacy `{ like: boolean }`.
Specify `{ like: true, caseSensitiveLike: true }` for case-sensitive search.
This change does not affect the underlying `$like` / `$ilike` implementation. 

TODOs:
- [X] MongoDB implementation
- [X] Generic SQL implementation (based on `Sequelize.fn('lower')`)
- [X] Use `ILIKE` operator for PostgreSQL

- [X] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No

Not a breaking change as custom endpoints are to be auto-migrated over.

### DIFF
--- a/modules/database/src/adapters/mongoose-adapter/MongooseSchema.ts
+++ b/modules/database/src/adapters/mongoose-adapter/MongooseSchema.ts
@@ -38,7 +38,7 @@ export class MongooseSchema implements SchemaAdapter<Model<any>> {
 
   async create(query: SingleDocQuery) {
     const parsedQuery = {
-      ...parseQuery(typeof query === 'string' ? EJSON.parse(query) : query),
+      ...(typeof query === 'string' ? EJSON.parse(query) : query),
       createdAt: new Date(),
       updatedAt: new Date(),
     };
@@ -46,7 +46,7 @@ export class MongooseSchema implements SchemaAdapter<Model<any>> {
   }
 
   async createMany(query: MultiDocQuery) {
-    const docs = parseQuery(typeof query === 'string' ? EJSON.parse(query) : query);
+    const docs = typeof query === 'string' ? EJSON.parse(query) : query;
     return this.model.insertMany(docs).then(r => r);
   }
 
--- a/modules/database/src/adapters/mongoose-adapter/MongooseSchema.ts
+++ b/modules/database/src/adapters/mongoose-adapter/MongooseSchema.ts
@@ -9,6 +9,7 @@ import {
   SingleDocQuery,
 } from '../../interfaces';
 import { MongooseAdapter } from './index';
+import { parseQuery } from './parser';
 import { ConduitSchema } from '@conduitplatform/grpc-sdk';
 import { isNil } from 'lodash';
 
@@ -37,7 +38,7 @@ export class MongooseSchema implements SchemaAdapter<Model<any>> {
 
   async create(query: SingleDocQuery) {
     const parsedQuery = {
-      ...(typeof query === 'string' ? EJSON.parse(query) : query),
+      ...parseQuery(typeof query === 'string' ? EJSON.parse(query) : query),
       createdAt: new Date(),
       updatedAt: new Date(),
     };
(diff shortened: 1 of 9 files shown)


modules/database/src/adapters/mongoose-adapter/MongooseSchema.ts
```diff
@@ -37,30 +38,22 @@ export class MongooseSchema implements SchemaAdapter<Model<any>> {
 
   async create(query: SingleDocQuery) {
     const parsedQuery = {
```
kkopanidis, line 40: why does create and createMany pass their docs through the parseQuery function? They do not contain operators

modules/database/src/adapters/mongoose-adapter/parser/index.ts
```diff
@@ -0,0 +1,58 @@
...
+        $and: query[key].map((operation: ParsedQuery) => {
+          return parseQuery(operation);
+        }),
+      });
+    } else {
+      if (!!query[key] && typeof query[key] === 'object' && !Array.isArray(query[key])) {
+        const likeCandidates = Object.keys(query[key]);
+        if (likeCandidates.includes('$like')) {
+          Object.assign(parsed, {
+            [key]: { $regex: `.*${query[key].$like.slice(1).slice(0, -1)}.*` },
```
kkopanidis, line 32: .* should not be applied automatically. like operations can have this string in either side or both of them, but that is a developer decision

modules/database/src/adapters/sequelize-adapter/parser/index.ts
```diff
@@ -93,23 +96,48 @@ function _parseQuery(
...
       });
     } else if (key === '$regex') {
       Object.assign(parsed, { [Op.regexp]: query[key] });
     } else if (key === '$options') {
       continue;
     } else {
-      const subQuery = _parseQuery(query[key], relations, associations);
+      if (!!query[key] && typeof query[key] === 'object' && !Array.isArray(query[key])) {
+        const likeCandidates = Object.keys(query[key]);
+        if (likeCandidates.includes('$like')) {
```
kkopanidis, line 115: are we expecting the "%" to be applied by the devs here?

modules/database/src/adapters/mongoose-adapter/parser/index.ts
kkopanidis: Basically a replace should be run to replace "%" with ".*" in the spots that the dev has specified.

kkopanidis (changes requested):

modules/database/src/adapters/mongoose-adapter/parser/index.ts
kon14: Adding a global replacement for '%'.
We should ideally handle escaped instances of '%' so that they're "escaped" instead of converted etc.

modules/database/src/adapters/sequelize-adapter/parser/index.ts
kon14: Yes. Custom endpoints automatically wrap the provided string so as to remain backwards compatible.
Direct query users should include percentage signs in their `$like` / `$ilike` operators.

modules/database/src/adapters/mongoose-adapter/parser/index.ts
```diff
@@ -0,0 +1,58 @@
+import { ParsedQuery } from '../../../interfaces';
+import { isArray, isBoolean, isNumber, isString } from 'lodash';
+import { Indexable } from '@conduitplatform/grpc-sdk';
+
+export function parseQuery(query: ParsedQuery): ParsedQuery {
+  return _parseQuery(query) as ParsedQuery;
+}
+
+export function _parseQuery(query?: ParsedQuery | null): ParsedQuery | null | undefined {
```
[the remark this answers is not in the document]
kon14: We could switch back to `.*` and forgo Mongoose query parsing entirely at this point..

kkopanidis (approved):

PR #547, opened 2023-03-02
feat: sql data types

SotiriaSte:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No

**Other information:**

### DIFF
--- a/libraries/grpc-sdk/src/interfaces/Model.ts
+++ b/libraries/grpc-sdk/src/interfaces/Model.ts
@@ -26,6 +26,9 @@ export enum SQLDataType {
   UUID = 'UUID',
   JSON = 'JSON',
 
+  // Postgres Specific
+
+  JSONB = 'JSONB',
   STRING = 'STRING',
 }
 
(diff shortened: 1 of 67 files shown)


kkopanidis (changes requested):
The SQLDataType should be in a new sqlType field an not extra options for the existing type field.

kon14 (commented):
While not necessarily breaking anything as unsupported types would simply fall back to the base type's default representation, shouldn't this come with a validation layer?

libraries/grpc-sdk/src/interfaces/Model.ts
```diff
@@ -8,6 +10,23 @@ export enum TYPE {
   Relation = 'Relation',
 }
 
+export enum SQLDataType {
```
kkopanidis, line 13: remove BLOB, JSONB, UUID and use TIME/DATETIME/TIMESTAMP values in extractType

kkopanidis (changes requested):

kkopanidis (approved):

PR #553, opened 2023-03-10
feat(database): sql PUT support

SotiriaSte:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No

**Other information:**


modules/database/src/adapters/sequelize-adapter/utils/schema.ts
```diff
@@ -136,12 +137,23 @@ export async function getTransactionAndParsedQuery(
...
+  if (method === 'PUT') {
+    const allFields = Object.keys(parsedQuery);
+    for (const field of Object.keys(parsedQuery)) {
+      if (allFields.indexOf(field) === -1) {
+        parsedQuery[field] = null;
+      }
+    }
+    parsedQuery = { $set: parsedQuery };
+  } else if (method === 'PATCH') {
+    if (parsedQuery.hasOwnProperty('$set')) {
```
kkopanidis, line 149: you need to check if the property is missing not if it is already there. If method is PATCH and parsedQuery has $set, then nothing is needed. If method is PATCH and $set is missing you add it. Also if method is PUT but $set exists then we just do patch. $set shouldn't be added on PUT

modules/database/src/adapters/sequelize-adapter/utils/schema.ts
```diff
@@ -136,12 +137,23 @@ export async function getTransactionAndParsedQuery(
   } else {
     parsedQuery = query;
   }
-  if (parsedQuery.hasOwnProperty('$set')) {
-    parsedQuery = parsedQuery['$set'];
+  if (method === 'PUT') {
+    const allFields = Object.keys(parsedQuery);
```
kkopanidis, line 141: parsedQuery doesn't contain allFields what is happening here?

kkopanidis (changes requested):
Also, the implementation is missing from mongo. Basically what mongo needs it "updateProvidedOnly" and if it is true add "$set", if "$set" already exists do nothing.

"updateProvidedOnly" is also missing from the protofile.

libraries/grpc-sdk/src/modules/database/index.ts
```diff
@@ -97,8 +97,12 @@ export class DatabaseProvider extends ConduitModule<typeof DatabaseProviderDefin
     });
   }
 
-  processQuery<T>(query: Query<T>) {
-    return JSON.stringify(query);
+  processQuery<T>(query: Query<T>, updateProvidedOnly = true) {
+    let parsedQuery = JSON.parse(JSON.stringify(query));
+    if (updateProvidedOnly && !parsedQuery.hasOwnProperty('$set')) {
```
kkopanidis, line 102: $set needs to be added by the adapters not by the sdk

modules/database/src/adapters/sequelize-adapter/utils/schema.ts
```diff
@@ -136,12 +137,20 @@ export async function getTransactionAndParsedQuery(
   } else {
     parsedQuery = query;
   }
-  if (parsedQuery.hasOwnProperty('$set')) {
+  if (method === 'PATCH') {
+    if (!parsedQuery.hasOwnProperty('$set')) {
```
kkopanidis, line 141: if it doesn't have $set you don't need to do anything here.

modules/database/src/adapters/sequelize-adapter/utils/schema.ts
```diff
@@ -136,12 +137,20 @@ export async function getTransactionAndParsedQuery(
...
+      parsedQuery = { $set: parsedQuery };
+    } else {
+      parsedQuery = parsedQuery['$set'];
+    }
+  }
+  if (method === 'PUT' && parsedQuery.hasOwnProperty('$set')) {
```
kkopanidis, line 147: Where did the code go that was setting null to fields that weren't sent?

kkopanidis (changes requested):

modules/database/src/adapters/sequelize-adapter/utils/schema.ts
SotiriaSte: > if it doesn't have $set you don't need to do anything here.

> you need to check if the property is missing not if it is already there. If method is PATCH and parsedQuery has $set, then nothing is needed. If method is PATCH and $set is missing you add it. Also if method is PUT but $set exists then we just do patch. $set shouldn't be added on PUT

kkopanidis: I got it mixed with mongo. $set doesn't do anything in sequelize so it shouldn't remain in the query. So to re-iterate: if method is PATCH and the query contains $set, you remove it and keep whatever is inside. If $set exists no matter the method we need to do "PATCH".

modules/database/src/adapters/sequelize-adapter/utils/schema.ts
```diff
@@ -136,12 +137,29 @@ export async function getTransactionAndParsedQuery(
...
+    } else {
+      parsedQuery = parsedQuery['$set'];
+    }
+  }
+  if (method === 'PUT') {
+    if (parsedQuery.hasOwnProperty('$set')) {
+      parsedQuery = parsedQuery['$set'];
+    } else {
+      const fields = Object.keys(parsedQuery);
+      for (const field of fields) {
```
kkopanidis, line 152: It's pointless to iterate on the fields passed. Your for loop should iterate on the fields that the schema we're trying to modify contains, and check if a field does not exist in the query, then set it to null

kkopanidis (changes requested):

PR #556, opened 2023-03-14
fix(database): some fixes for import & export

ChrisPdgn:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/database/src/admin/schema.admin.ts
+++ b/modules/database/src/admin/schema.admin.ts
@@ -35,7 +35,10 @@ export class SchemaAdmin {
         undefined,
         undefined,
         'name parentSchema fields extensions modelOptions ownerModule collectionName',
-      );
+      )
+      .then(r => {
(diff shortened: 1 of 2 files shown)


modules/database/src/admin/schema.admin.ts
```diff
@@ -35,7 +35,11 @@ export class SchemaAdmin {
         undefined,
         undefined,
         'name parentSchema fields extensions modelOptions ownerModule collectionName',
-      );
+        { updatedAt: 1 },
```
kon14, line 38: This won't fix import requests hanging due to yet-to-be-imported related schemas being sorted incorrectly.
We should figure out why the pending schema functionality won't work on the import side.

kkopanidis (approved):

PR #605, opened 2023-04-24
fix(module-tools,storage):  gRPC int64 long types

kon14:
This PR fixes int64 types being typed as `number` without getting implicitly decoded from `string` long repr.
These proto fields are now treated as `protobufjs::Long` instead and explicitly decoded as required.
This resolves `Storage`'s `createFileByUrl`/`updateFileByUrl` `size` fields causing a db create operation crash due to invalid data types in cases where `size` does not get auto-cast.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No

### DIFF
--- a/libraries/grpc-sdk/build.sh
+++ b/libraries/grpc-sdk/build.sh
@@ -13,6 +13,7 @@ protoc \
   --ts_proto_opt=esModuleInterop=true \
   --ts_proto_opt=outputServices=generic-definitions,exportCommonSymbols=false,useExactTypes=false \
   --ts_proto_out=./ \
+  --ts_proto_opt=forceLong=long \
   ./*.proto
 
 # Initialize the content of the index.ts file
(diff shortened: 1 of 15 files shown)


libraries/grpc-sdk/build.sh
```diff
@@ -13,6 +13,7 @@ protoc \
   --ts_proto_opt=esModuleInterop=true \
   --ts_proto_opt=outputServices=generic-definitions,exportCommonSymbols=false,useExactTypes=false \
   --ts_proto_out=./ \
+  --ts_proto_opt=forceLong=long \
```
kkopanidis, line 16: remove this is already the default option

kkopanidis (changes requested):

libraries/grpc-sdk/build.sh
kon14: According to [ts-proto's repo docs](https://github.com/stephenh/ts-proto)

`The default behavior is forceLong=number, which will internally still use the long library to encode/decode values on the wire (so you will still see a util.Long = Long line in your output), but will convert the long values to number automatically for you. Note that a runtime error is thrown if, while doing this conversion, a 64-bit value is larger than can be correctly stored as a number.`

Default option generates `number` types.

PR #608, opened 2023-04-25
fix(database): support $in & $nin operators for all sql dialects

ChrisPdgn:
This PR is an attempt to provide support for $in & $nin operators for all sql dialects and all possible cases of use.

Also, a fix is on the way for a population issue that causes the $nin queries on relation array columns to return wrong results.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No


ChrisPdgn:
I have just found in docs that the proper form of this query in order to work is this:
```javascript
  const result = await Test.findAll({
    where: {
      [Op.not]: {
        verbs: {
          [Op.overlap]: ['get', 'set']
        }
      }
    }
  });
```
Therefore, this issue was my mistake

kon14 (approved):

PR #619, opened 2023-05-06
Type rework

kkopanidis:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

**Other information:**


modules/database/src/adapters/utils/validateFieldConstraints.ts
```diff
@@ -1,16 +1,36 @@
...
+  schemaFields: ConduitModel,
+  db: string,
+  blockRelations = false,
+) {
   Object.keys(schemaFields).forEach(f => {
+    if (/(?<=[a-zA-Z0-9])_(?=[a-zA-Z0-9])|(?<=[a-zA-Z0-9])_$/g.test(f)) {
+      throw new ConduitError(
+        'INVALID_ARGUMENTS',
+        400,
+        `Schema '${schemaName}' violates field '${f}' constraint (field names cannot contain '_' in-between characters).`,
```
kon14, line 24: What if we used double underscores instead?
That would preserve support for more field names (probably not uncommon for foreign dbs, also digit words), while also retaining indexOf() checks.

modules/router/src/admin/index.ts
```diff
@@ -100,12 +100,9 @@ export class AdminHandlers {
           path: ConduitString.Required,
           target: ConduitString.Required,
           action: ConduitString.Required,
-          description: ConduitString.Optional,
+          routeDescription: ConduitString.Optional,
```
kon14, line 103: But why?

libraries/module-tools/src/helpers/TypeHelpers.ts
```diff
@@ -1,14 +1,14 @@
-import { ConduitModelField, TYPE } from '@conduitplatform/grpc-sdk';
+import { TYPE } from '@conduitplatform/grpc-sdk';
 
 class ConduitStringConstructor {
   // private to disallow creating other instances of this type
   private constructor() {}
 
-  static get Optional() {
+  static get Optional(): TYPE.String {
```
kon14, line 7: Why so verbose?

modules/database/src/adapters/mongoose-adapter/MongooseSchema.ts
```diff
@@ -10,7 +10,7 @@ import {
 } from '../../interfaces';
 import { MongooseAdapter } from './index';
 import { parseQuery } from './parser';
-import { ConduitSchema } from '@conduitplatform/grpc-sdk';
+import { ConduitModel, ConduitSchema, Indexable } from '@conduitplatform/grpc-sdk';
```
kon14, line 13: Unused import

kon14 (changes requested):
Shouldn't we also prohibit 'type' fields in schemas?

kkopanidis:
> Shouldn't we also prohibit 'type' fields in schemas?

we do

modules/database/src/adapters/utils/validateFieldConstraints.ts
kkopanidis: Double underscores are not common practice for signifying "objects" in tables, and it'd be very ugly in practice. The only other applicable solution would be to do something smarter in wrap/unwrap ops that wouldn't confuse random underscores as the existence of an object.

kkopanidis: Also this is a db rework issue, and your concerns should be addressed at the other PR

modules/router/src/admin/index.ts
kkopanidis: Because it's a ConduitModel, and it disallows the use of "description" as a reserved word.

libraries/module-tools/src/helpers/TypeHelpers.ts
kkopanidis: The TS compiler was having issues

ChrisPdgn (approved):

PR #657, opened 2023-07-10
refactor(database,authorization,grpc-sdk): Mongo authz query aggregations

ChrisPdgn:
This PR refactors the way we use access control in mongoose. Now, authorization module doesn't create views but returns authorized queries that calculate the accessible docs through aggregation. 

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**


kon14 (approved):

kon14:
This seems to be working as expected.

We should ideally delete any `View` docs for Mongo users, but given how the gRPC API already exposed view creation we could be deleting non-authz ones explicitly created by the users themselves.

modules/authorization/src/utils/index.ts
```diff
@@ -31,56 +31,128 @@ export const computePermissionTuple = (
...
+            SELECT * FROM cnd_ActorIndex
+            WHERE subject = '${subject}'
+        ) actors ON 1=1
+        INNER JOIN (
+            SELECT * FROM cnd_ObjectIndex
+            WHERE subject LIKE '${objectType}:%#${action}'
+        ) objects ON actors.entity = objects.entity;`;
+}
+
+export function getMongoAccessListQuery(params: AccessListQueryParams) {
```
ChrisPdgn, line 76: I think this is the query we used to have a long time ago because it doesn't seem to match with the one deleted in createAccessList. Please check and update it we the new one we have if I'm correct

PR #681, opened 2023-09-08
feat(authorization, database):  createRelations()

Renc17:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/authorization/src/Authorization.ts
+++ b/modules/authorization/src/Authorization.ts
@@ -1,4 +1,5 @@
 import ConduitGrpcSdk, {
+  CreateRelationsRequest,
   DatabaseProvider,
   GrpcRequest,
   GrpcResponse,
@@ -45,6 +46,7 @@ export default class Authorization extends ManagedModule<Config> {
       deleteResource: this.deleteResource.bind(this),
       updateResource: this.updateResource.bind(this),
       createRelation: this.createRelation.bind(this),
+      createRelations: this.createRelations.bind(this),
       grantPermission: this.grantPermission.bind(this),
       removePermission: this.removePermission.bind(this),
       deleteRelation: this.deleteRelation.bind(this),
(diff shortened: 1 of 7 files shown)


modules/authorization/src/admin/relations.ts
```diff
@@ -20,7 +20,7 @@ export class RelationHandler {
   registerRoutes(routingManager: RoutingManager) {
     routingManager.route(
       {
-        path: '/relations',
+        path: '/relation',
```
kon14, line 23: This is breaking the API and convention. Make this `/relations/` and set the createMany variant to `/relations/many` instead.

modules/authorization/src/controllers/relations.controller.ts
```diff
@@ -58,6 +58,69 @@ export class RelationsController {
     return relationResource;
   }
 
+  async checkRelations(subject: string, relation: string, resources: string[]) {
+    checkRelation(subject, relation, resources[0]);
+    const computedTuples = resources.map(r => computeRelationTuple(subject, relation, r));
+    const relationResources = await Relationship.getInstance().findMany({
+      computedTuple: { $in: computedTuples },
+    });
+    if (relationResources.length) throw new Error('Relations already exists');
```
kon14, line 67: This is probably gonna get rebased anyway, but simply ignore existing relations instead.
If you'd rather stick with whatever's already in `main`, fix singular/plural typo.

kon14 (changes requested):

modules/authorization/src/Authorization.ts
```diff
@@ -125,6 +127,18 @@ export default class Authorization extends ManagedModule<Config> {
...
   }
 
+  async createRelations(
+    call: GrpcRequest<CreateRelationsRequest>,
+    callback: GrpcResponse<Empty>,
+  ) {
+    const subject = call.request.subject;
+    const relation = call.request.relation;
+    const resources = call.request.resources;
+    await this.relationsController.checkRelations(subject, relation, resources);
```
kon14, line 137: Why is this explicitly called outside `createRelations()` in the first place?
Doesn't seem to be used anywhere else either.
Imo, make it private or integrate directly back into `createRelations()`.

kon14 (changes requested):

kon14 (approved):

PR #683, opened 2023-09-11
feat(database): add import/export extensions logic to existing endpoints

ChrisPdgn:
Import/export endpoints now allow the importing/exporting of conduit schemas' extensions
Also, this PR fixes a bug in extension setting

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/modules/database/src/admin/schema.admin.ts
+++ b/modules/database/src/admin/schema.admin.ts
@@ -26,8 +26,8 @@ export class SchemaAdmin {
     private readonly customEndpointController: CustomEndpointController,
   ) {}
 
-  async exportCustomSchemas(): Promise<UnparsedRouterResponse> {
-    return await this.database
+  async exportSchemas(): Promise<UnparsedRouterResponse> {
+    const cmsSchemas = await this.database
       .getSchemaModel('_DeclaredSchema')
       .model.findMany(
         { 'modelOptions.conduit.cms.enabled': true },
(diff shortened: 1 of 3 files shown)


kkopanidis (approved):

modules/database/src/admin/schema.admin.ts
```diff
@@ -38,45 +38,57 @@ export class SchemaAdmin {
...
-        ),
-        operation,
-        imported,
-      );
       if (schema.extensions.length > 0) {
         for (const extension of schema.extensions) {
           await this.database.setSchemaExtension(
             schema.name,
-            extension.owner,
+            extension.ownerModule,
```
kon14, line 91: This is technically breaking the `API` (`owner` -> `ownerModule`).
But it seems like the export endpoint actually returns `ownerModule`.
As such, I'm fine with this, but let's document in PR (include breaking change in conventional commit)

ChrisPdgn: This isn't a breaking change, because as you said, the export returns the extensions as they are stored in a schemas (``ownerModule``), meaning the result of export doesn't change at all. It would be a breaking change if the extensions used to be stored in schemas with ``owner`` (but as I checked, they were always ``ownerModule`` so this is just a bug fix)

modules/database/src/admin/schema.admin.ts
```diff
@@ -38,45 +38,57 @@ export class SchemaAdmin {
...
-      const modelOptions = SchemaConverter.getModelOptions({
-        cmsSchema: true,
-        existingModelOptions: schema.modelOptions,
-        importedSchema: imported,
-      });
-      try {
-        validateSchemaInput(schema.name, schema.fields, modelOptions);
-      } catch (err: unknown) {
-        throw new GrpcError(status.INTERNAL, (err as Error).message);
+      const isCmsSchema = schema.modelOptions;
```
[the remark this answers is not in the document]
kon14: No, you're right about `modelOptions` (righfully) not being exported.
I read that part wrong (confused with db schema entries.

This technically works as it is, but we should be crosschecking whether imported schemas are actually module schemas or not regardless of what's provided in the request body.

What if I were to provide `modelOptions` in a module schema?
We're currently treating schema names as unique, therefore we could compare with the ones available in the adapter. If unknown or `schema.modelOptions.cms`, treat as CMS schema, otherwise throw early with an appropriate error message.
These checks would ideally take place before any modifications have taken place.

modules/database/src/admin/schema.admin.ts
kon14: It's a breaking change regardless of whether it's a fix or not.
API behaved in one way, now it doesn't anymore.

But this isn't really a front-facing endpoint, so we shouldn't worry too much.

modules/database/src/admin/schema.admin.ts
```diff
@@ -38,45 +38,57 @@ export class SchemaAdmin {
...
-        new ConduitSchema(
-          schema.name,
-          schema.fields,
-          modelOptions,
-          schema.collectionName,
-        ),
-        operation,
-        imported,
-      );
       if (schema.extensions.length > 0) {
```
kon14, line 87: This doesn't allow for removing extensions.
This endpoint receives a fairly complete representation of the included schemas (save for indexes), therefore missing extensions should be unset.

The only case were dropping extensions would feel wrong is when said extensions are owned by foreign modules. These should most likely be setting their own extensions through gRPC.
As such, I think we should only be setting/unsetting database-owned extensions and either throwing (early) or ignoring any module-specific ones.

kon14 (changes requested):

modules/database/src/admin/schema.admin.ts
```diff
@@ -42,21 +47,38 @@ export class SchemaAdmin {
...
   async importSchemas(call: ParsedRouterRequest): Promise<UnparsedRouterResponse> {
     const schemas = call.request.params.schemas;
     for (const schema of schemas) {
-      const isCmsSchema = schema.modelOptions;
+      const found = await this.database
+        .getSchemaModel('_DeclaredSchema')
+        .model.findOne({ name: schema.name });
+      const isCmsSchema = found
+        ? !!found.modelOptions.conduit.cms
+        : !!schema?.modelOptions.conduit.cms;
```
kon14, line 81: Right-hand case should be `!!schema.modelOptions?.conduit?.cms` (or similar).
Schema is always gonna be defined as you're looping through the schema array, but any nested object fields could (and technically should) be missing for module schemas.

kon14 (changes requested):
I have yet to retest exporting/importing at the latest commit.
This is specially regarding the cms check.

modules/database/src/admin/schema.admin.ts
```diff
@@ -38,49 +43,75 @@ export class SchemaAdmin {
...
-          await this.database.setSchemaExtension(
-            schema.name,
-            extension.owner,
-            extension.fields,
-          );
+      const isCmsSchema = found
+        ? !!found.modelOptions.conduit.cms
+        : !!schema.modelOptions?.conduit?.cms;
+      if (isCmsSchema) {
+        const existingSchema = await this.database
```
kon14, line 81: There's no reason why this should refetch the schema from the database.
In fact, we should ideally be performing a `findMany()` above too.

modules/database/src/admin/schema.admin.ts
```diff
@@ -38,49 +43,75 @@ export class SchemaAdmin {
...
+          .getSchemaModel('_DeclaredSchema')
+          .model.findOne({ name: schema.name });
+        const operation = isNil(existingSchema) ? 'create' : 'update';
+        const imported = schema.modelOptions.conduit.imported;
```
kon14, line 85: Prefer db-fetched schema's imported value, falling back to arg-provided one if not an existing schema.

modules/database/src/admin/schema.admin.ts
```diff
@@ -38,49 +43,75 @@ export class SchemaAdmin {
...
+            schema.collectionName,
+          ),
+          operation,
+          imported,
+        );
       }
+      const extension = (schema.extensions as DeclaredSchemaExtension[]).find(
+        ext => ext.ownerModule === 'database',
+      );
+      await this.database.setSchemaExtension(
```
kon14, line 110: This would crash for non-cms schemas that have yet to be created.
We should either:
a) check these beforehand and crash early
b) keep track of missing module schemas and alter the endpoint's response from `string` to `{ status: string, missing: string[] }` (or similar), where status would indicate success or partial success.

Let's discuss this before implementation.

kon14 (changes requested):
Additionally, export schemas endpoint should either:
- completely ignore non-cms schemas without extensions (don't list)
- only list non-cms schemas with `modelOptions.conduit.permissions.extendable`

kon14 (approved):

PR #764, opened 2023-10-24
feat(chat): add config to delete rooms when participants have left

kkopanidis:
One extra idea, that we should probably think about implementing, is the "auditing" of the chats. This could be enabled optionally, so when a chat is deleted, then it would only be flagged as deleted and not removed from the database. It should also keep track of who and when, entered or left the chat, along with their messages, so that an admin can audit the messages for abuse.

Thinking even more long-term, even if end-to-end encryption was introduced in an app, the user could be able to share encryption keys in order read the actual messages for such cases.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/modules/chat/src/config/config.ts
+++ b/modules/chat/src/config/config.ts
@@ -11,7 +11,8 @@ export default {
     format: 'Boolean',
     default: true,
   },
-  destroyRoomWhenEmpty: {
+  deleteEmptyRooms: {
+    doc: 'Defines whether rooms should be automatically deleted after all participants have left',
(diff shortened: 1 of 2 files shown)


kon14 (approved):

kon14:
We should technically also be handling User deletion events and propagating the effects here.
Our bus event handling is currently inconsistent at best though. Opening a task.

PR #891, opened 2024-01-14
refactor(authorization): rework

kkopanidis:
This PR reworks the index creation of the authorization module. The target is to maintain the current behaviour of constructing inherited permissions that are deeply nested, without requiring the use of background jobs and greedy searching.

This is achieved by introducing 2 specific changes:
1. When creating indexes for direct relations, the module also creates an object index for the subject's relations, according to the inheritance rules of the object.
2. Combined with 1, a new field is added on object indexes, which grant permissions based on inheritance, which specifies which specific permission of the subject is getting used to grant a permission X on the object. This allows us to lookup these indexes when computing deep-nested relations, without requiring to find the related subjects first.

This approach, while it fixes the greedy computations, it introduces more space complexity for the DB. But, I haven't been able to find, although I've searched far and wide, a solution that handles space more efficiently while preserving the ridiculous fast querying times we have. Using a dedicated db, in the future, like neo4j which handle graphs natively, would allow us to remove the indexes, but would have the drawback of 1. requiring yet another thing to be deployed and 2. instead of sending pre-computed queries to the DB to filter out non-authorized results, to send back IDs of matching objects, which would introduce new problems entirely.

TODO before merge:
- [ ] Check re-indexing performance
- [ ] Check and possible improve, bulk relation performance
- [ ] Check resource modification behaviour to make sure no permissions "leak"
- [ ] Check permission removal behaviour to make sure no permissions "leak"
- [ ] Evaluate overall validity of queries and performance on large datasets

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**


kon14 (commented):
`PermissionsController.createAccessList` doesn't seem to support `*` permission roles.

kkopanidis:
@kon14  issue resolved

PR #922, opened 2024-01-26
fix: standalone Docker compose port mapping

kon14:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [X] No


docker/docker-compose.standalone.yml
```diff
@@ -19,8 +19,20 @@ services:
       - mongodb
     ports:
       - '${CORE_GRPC_PORT:-55152}:${CORE_GRPC_PORT:-55152}'
-      - '${CORE_ADMIN_HTTP_PORT:-3030}:3030'
-      - '${CORE_ADMIN_SOCKET_PORT:-3031}:3031'
+      - '${DB_GRPC_PORT:-55160}:${DB_GRPC_PORT:-55160}'
```
kkopanidis, line 22: we cannot allow modification of module ports in this setup, since the modules won't change the port they are listening to. We CAN however change the mapping, so the right part of the port assignment needs to be fixed, but the left can be modular. Even then though, you would need to use re-mapping in the sdk for the requests to work, since modules will report their defined port and not the port the user added to the env file.

kkopanidis (changes requested):

docker/docker-compose.standalone.yml
kon14: Even without SDK re-mapping, host port mapping overrides could still prove useful in cases where the internal port is not available on the host and users wish to interact with the services on the host (for SDK-less API calls and/or testing).

kkopanidis (approved):

PR #1091, opened 2024-07-08
feat(email): store emails to db/storage

ChrisPdgn:
This PR introduces some new email module features: 
- Store sent emails to database or storage (includes cleanup job). Emails stored in DB contain only some needed information of the sent email for debugging reasons, whereas emails stored in storage contain additionally the full context
- Resend a stored email (only if stored in storage)
- Get the latest status of a sent email (for this to work, please make sure your provider account has full access to email activity history)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**


modules/email/src/services/email.service.ts
```diff
@@ -54,7 +61,12 @@ export class EmailService {
     });
   }
 
-  async sendEmail(template: string, params: ISendEmailParams) {
+  async sendEmail(
```
kkopanidis, line 64: this function already supports sending emails with pre-compiled body, by providing the body param in the params object. To do that you need to set template to null. Obviously that is currently not supported through typescript since template is string. Changing template type to string | undefined will allow you to provide the body and subject params without issue.

modules/email/src/services/email.service.ts
```diff
@@ -128,6 +142,55 @@ export class EmailService {
     if (params.attachments) {
       builder.addAttachments(params.attachments as Attachment[]);
     }
-    return this.emailer.sendEmail(builder);
+
+    const sentMessageInfo = await this.emailer.sendEmail(builder);
```
kkopanidis, line 146: I would advise to disable creating a new storage entry if the email is a resend. That can work by either creating a database record that points to the existing email content, or by updating the existing record

kkopanidis (changes requested):

kkopanidis (approved):

Issue #1076, opened 2024-06-14
[FEAT]: Support for Files Alias Name


Renc17:
### Suggestion

I propose adding a new feature to the storage module that allows users to specify an alias name for uploaded files. This functionality will enable users to provide a custom name for each file upon upload, addressing potential duplication issues when files with identical names are uploaded.

To achieve this, I propose we use a generated UUID instead of indexing file names for uniqueness.

kkopanidis:
My thoughts on implementing this are as follows:
- Make file name optional in file requests. If no name is provided a UUID will be create for file name
- Introduce a new field for requests and the db, where you can specify the name of the file as you want it to appear, but not as you want it stored.
Pros:
Not breaking, allows for better visualization in the client side
Cons:
Maybe a bit more complex to understand

Alternatively, we could change the usage of the name field and always use it as an "alias" but not as the way that the file is going to be stored. So files will always be stored with UUIDs, but will appear with the specified name in the db.
Pros:
Straightforward approach in handling
Cons:
Potentially breaking change, cannot identify file if you're exploring the storage filesystem, don't have the option to specify the name in the storage.

Thoughts @Renc17, @ChrisPdgn, @kon14 ?

kkopanidis:
@ChrisPdgn that's another behaviour that we could look into. In general uniqueness is enforced db-wise I think, can't remember precisely. So a file with name X needs to be unique across everything. But even if we consider container/bucket and folder names, this would work nicely. I don't know if it should be the default behaviour though, or if the strategy should be specified on the settings and/or the request to create the file. Meaning to pick between setting (1), (2) etc.  for example or use the uuid to store the file and keep the name in the db

kkopanidis:
done in #1096

PR #1096, opened 2024-07-26
feat(storage): file aliases

ChrisPdgn:
This PR implements some storage module features mentioned in #1076 
The ``name`` field in storage file requests in now optional & another optional field ``alias`` is introduced. 
This enables creating files with a same alias while storing them with a UUID name

Additionally, a configurable storage feature is introduced - appending a suffix at the end of a name in case of conflict.
This feature is mimicking the way files are traditionally stored in file systems

Finally, while adding the features I did some refactoring to minimize found duplicate code

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/modules/storage/src/utils/index.ts
+++ b/modules/storage/src/utils/index.ts
@@ -3,8 +3,10 @@ import { IFileParams, IStorageProvider, StorageConfig } from '../interfaces/inde
 import { isNil } from 'lodash-es';
 import path from 'path';
 import { File } from '../models/index.js';
-import { ConduitGrpcSdk } from '@conduitplatform/grpc-sdk';
+import { ConduitGrpcSdk, GrpcError } from '@conduitplatform/grpc-sdk';
 import { randomUUID } from 'node:crypto';
+import { ConfigController } from '@conduitplatform/module-tools';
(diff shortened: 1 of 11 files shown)


modules/storage/src/utils/index.ts
```diff
@@ -62,3 +67,176 @@ export async function deepPathHandler(
...
+      { folder: folder },
+      { container: container },
+    ],
+  });
+  if (count === 0) {
+    return name;
+  } else if (!config.suffixOnNameConflict) {
+    throw new GrpcError(status.ALREADY_EXISTS, 'File already exists');
+  } else {
+    return `${name} (${count})`;
```
kkopanidis, line 240: count needs to be added before file type, if file type exists in name. For example on file myFile.pdf it should become myFile (1).pdf. While myFile, should become myFile (1).

kkopanidis (changes requested):
For the life of me I can't find the UUID generation in the code or any check for name nullity

kkopanidis (approved):

PR #1100, opened 2024-08-06
refactor(push-notifications): rework some admin endpoints for the new UI and correctness

kkopanidis:
feat(push-notifications): set default state to enabled

In truth the endpoint changes are breaking, but they are not in use really, so I'm kinda cheating here.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/modules/push-notifications/src/admin/index.ts
+++ b/modules/push-notifications/src/admin/index.ts
@@ -234,7 +234,7 @@ export class AdminHandlers {
           skip: ConduitNumber.Optional,
           limit: ConduitNumber.Optional,
           sort: ConduitString.Optional,
-          search: ConduitString.Optional,
+          search: ConduitObjectId.Optional,
           platform: ConduitString.Optional,
           populate: ConduitString.Optional,
         },
--- a/modules/push-notifications/src/admin/index.ts
+++ b/modules/push-notifications/src/admin/index.ts
@@ -3,6 +3,7 @@ import {
   ConduitRouteActions,
   ConduitRouteReturnDefinition,
   GrpcError,
+  Indexable,
   ParsedRouterRequest,
   TYPE,
   UnparsedRouterResponse,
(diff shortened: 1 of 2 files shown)


modules/push-notifications/src/admin/index.ts
```diff
@@ -172,17 +227,55 @@ export class AdminHandlers {
...
       {
-        path: '/token/:userId',
+        path: '/token',
+        action: ConduitRouteActions.GET,
+        description: `Get and search notification tokens.`,
+        queryParams: {
+          skip: ConduitNumber.Optional,
+          limit: ConduitNumber.Optional,
+          sort: ConduitString.Optional,
+          search: ConduitString.Optional,
```
Renc17, line 237: this should be typeof ConduitObjectId

Renc17 (changes requested):

modules/push-notifications/src/admin/index.ts
```diff
@@ -97,16 +100,68 @@ export class AdminHandlers {
...
 
+  async getNotificationTokenByUserId(
+    call: ParsedRouterRequest,
+  ): Promise<UnparsedRouterResponse> {
+    const tokenDocuments = await NotificationToken.getInstance().findMany(
+      {
+        userId: call.request.urlParams.userId,
+      },
+      undefined,
+      undefined,
```
Renc17, line 111: what if user has more tokens than the default limit of findMany? Shouldn't we add pagination here or at least set sorting by descending creation date which I think would be the most common case.

kkopanidis: skip/limit defaults are not applied to rpc functions

Renc17 (approved):

PR #1115, opened 2024-08-27
refactor(grpc-sdk): sdk function params as object option

ChrisPdgn:
This PR adds the option to provide the params of some sdk functions as an object to avoid setting them as ``undefined``. The old way of function calling is still supported

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**


libraries/grpc-sdk/src/utilities/normalizeParams.ts
```diff
@@ -0,0 +1,10 @@
+export function normalizeParams(givenParams: any[], fnParamKeys: string[]) {
```
kon14, line 1: `givenParams`'s type structure feels ambiguous and somewhat error-prone.
Object variant shouldn't be array-nested (nor should the actual function type unions wrap the object subtypes in an array).

This makes it possible for an array of objects to be passed along instead of a plain object value.
This skews ergonomics, in that you'd need to array-wrap the actual values during the call, but more importantly introduces an unchecked flow.
What happens if an array of object-formatted args is passed in?

libraries/grpc-sdk/src/utilities/normalizeParams.ts
```diff
@@ -0,0 +1,10 @@
...
+  if (givenParams.length === 1 && typeof givenParams[0] === 'object') {
```
kon14, line 2: What if the array-variant expects an object arg at pos[0]?
This would only pass in the first arg, assuming object formatting.

kon14 (changes requested):

libraries/grpc-sdk/src/utilities/normalizeParams.ts
ChrisPdgn: I changed the logic, now only optional params can be set as an object

kkopanidis (approved):

PR #1160, opened 2024-09-18
feat(comms)!: new communications module that merges SMS, Push and Email into a single service

### LINKED ISSUE #1172: [FEAT] Merge Push, Email & SMS modules
### Discussed in https://github.com/ConduitPlatform/Conduit/discussions/122

<div type='discussions-op-text'>

<sup>Originally posted by **kkopanidis** February  8, 2022</sup>
Currently conduit spins up multiple containers that don't really make sense to be separate, since they can't practically scale independently. Email, push and sms utilise different external providers for their operations, but they fulfil a common goal of outgoing operations with templates etc.

I think it'd be best for these to be merged into a singular "communications" module, that handles all outgoing user communications with either channel. This will allow for even cool things like "reach the user by any mean necessary", and incorporate more channels ex. viber, whatsapp etc.

**Additional context**
This particular module merge doesn't have any scalability arguments i think, since basically their capacity dependes 80-90% to the external services and merging them shouldn't provide any significant benefits or drawbacks on that part. But, it'd definitely save up some ram and cpu in the deployments.
</div>

kkopanidis:
This PR implements #122 merging the SMS, Push Notifications and Email modules into a single Comms module. The aim of this merge is to:
- Reduce the required memory of the modules
- Simplify operations
- Allow for increased features like communication preferences and easier user notifications without channel picks

Tasks:
- [x] Merge modules
- [x] Migrations for configuration
- [x] Migrations for schemas
- [x] Migration for admin routes
- [x] Migrate current usage of sub-services in other modules 

This is a breaking change, to be merged in v-next branch for version 0.17.x

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

**Other information:**


modules/comms/src/Comms.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  async initializeMetrics() {
+    await this.smsService.initializeMetrics?.();
+    await this.emailService.initializeMetrics?.();
+    await this.pushService.initializeMetrics?.();
+  }
+
+  // gRPC Service
+  async featureAvailable(
```
kon14, line 128: Delivering this RPC method goes against everything I know to be right and true and I will sooner lay you into this barren earth, than entertain your folly for a moment longer. [[1](https://www.youtube.com/watch?v=y8OnoxKotPQ)]

Jokes aside, this pointlessly breaks [gRPC Health Checking Protocol v1](https://github.com/grpc/grpc/blob/master/doc/health-checking.md):

> The client can optionally set the service name it wants to query for health status.
>
> The server should register all the services manually and set the individual status, including an empty service name and its status. For each request received, if the service name can be found in the registry, a response must be sent back with an OK status and the status field should be set to SERVING or NOT_SERVING accordingly. If the service name is not registered, the server returns a NOT_FOUND GRPC status.
>
> The server should use an empty string as the key for server's overall health status, so that a client not interested in a specific service can query the server's status with an empty request. The server can just do exact matching of the service name without support of any kind of wildcard matching. However, the service owner has the freedom to implement more complicated matching semantics that both the client and server agree upon.

``` proto
message HealthCheckRequest {
  string service = 1;
}
```

All that's needed for compatibility to be preserved is for `Comms` to integrate this pattern matching in the context of health `check()`/`watch()`.

While`ConduitServiceModule` doesn't currently explicitly accommodate this functionality, merely overriding the base methods wouldn't cause any additional breaking changes, nor overcomplicate the `module-tools` API for single service consumers.
Here's the current impl for reference:

``` ts
healthCheck(
  call: GrpcRequest<HealthCheckRequest>,
  callback: GrpcCallback<HealthCheckResponse>,
) {
  const service = call.request.service.substring(call.request.service.indexOf('.') + 1);
  if (service && service !== this._serviceName) {
    callback(null, {
      status:
        HealthCheckStatus.SERVICE_UNKNOWN as unknown as HealthCheckResponse_ServingStatus,
    });
  } else {
    if (!this._registered) {
      return callback(null, {
        status: HealthCheckResponse_ServingStatus.UNKNOWN,
      });
    }
    callback(null, {
      status: this._serviceHealthState as unknown as HealthCheckResponse_ServingStatus,
    });
  }
}

healthWatch(call: ServerWritableStream<HealthCheckRequest, HealthCheckResponse>) {
  const service = call.request.service.substring(call.request.service.indexOf('.') + 1);
  if (service && service !== this._serviceName) {
    call.write({
      status:
        HealthCheckStatus.SERVICE_UNKNOWN as unknown as HealthCheckResponse_ServingStatus,
    });
  } else {
    this.events.on(
      `grpc-health-change:${this._serviceName}`,
      (status: HealthCheckStatus) => {
        call.write({
          status: status as unknown as HealthCheckResponse_ServingStatus,
        });
      },
    );
  }
}
```

kon14 (changes requested):

modules/comms/src/Comms.ts
kkopanidis: @kon14 I agree, but this can't be part of this PR to be honest. I see 2 possibilities, either we make changes for health-checking utilities in another PR, merge it and apply changes here as well, or we merge this as-is, change the health checking and apply changes afterwards. 

I prefer no2 because I am doing a re-work of health checking from the core side, in order for it to work properly with dns names on platforms like k8s or docker compose when DNS is involved, so in the same PR I can tackle sdk/module-tools checks for multi-service modules as well. If you agree with that, open it as an issue and I'll add it to the roadmap (I'm preparing the 0.17 roadmap project)

PR #1184, opened 2024-10-08
feat(authentication): invitations & additional user data in registration

ChrisPdgn:
This PR implements some needed features for user registration & team invitations
- Team invitations now automatically verify a user's email (if email mismatch is allowed)
- Team invitation token data can now include some ``userData``, so that additional info can be stored before a user registers
- ``GET /teams/invite/user/data`` endpoint to retrieve the data mentioned above
- New persistent invitation logic with ``POST, DELETE /teams/:teamId/invite/persistent`` that allows a team owner to invite multiple users to a team with a non-expiring token & without setting user emails
- Optional ``userData`` JSON param in ``POST /local`` & ``POST /phone/verify`` allows adding fields to user doc to avoid calling multiple endpoints for user registration.
In case of OAuth, this param cannot/should not be included as these methods are the first step of registration. The automatically verified user can later call the new ``PATCH /user`` endpoint to set additional fields

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/modules/authentication/src/handlers/common.ts
+++ b/modules/authentication/src/handlers/common.ts
@@ -6,6 +6,7 @@ import {
   ConduitRouteReturnDefinition,
   GrpcError,
   ParsedRouterRequest,
+  TYPE,
   UnparsedRouterResponse,
 } from '@conduitplatform/grpc-sdk';
 import { status } from '@grpc/grpc-js';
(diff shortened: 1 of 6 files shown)


modules/authentication/src/handlers/common.ts
```diff
@@ -150,6 +159,24 @@ export class CommonHandlers implements IAuthenticationStrategy {
       new ConduitRouteReturnDefinition('DeleteUserResponse', 'String'),
       this.deleteUser.bind(this),
     );
+    routingManager.route(
+      {
+        path: '/user',
+        description: `Updates user.`,
+        action: ConduitRouteActions.PATCH,
+        bodyParams: {
+          fields: {
```
kkopanidis, line 168: rename to userData so that we don't get any naming conflicts

modules/authentication/src/handlers/common.ts
```diff
@@ -129,6 +130,14 @@ export class CommonHandlers implements IAuthenticationStrategy {
     return this.logOut(call);
   }
 
+  async updateUser(call: ParsedRouterRequest): Promise<UnparsedRouterResponse> {
+    const { user } = call.request.context;
+    const fields = call.request.params.fields;
```
kkopanidis, line 135: fetch the data from the body specifically, don't rely on params (we should have deprecated that tbh)

PR #1333, opened 2025-04-28
feat(email): email status polling

krigol14:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

Links to the specific documentation section of each provider:
[Mailgun](https://documentation.mailgun.com/docs/mailgun/api-reference/openapi-final/tag/Events/)
[Mailersend](https://developers.mailersend.com/general.html#email-resource)
[Sendgrid](https://www.twilio.com/docs/sendgrid/api-reference/email-activity/filter-messages-by-message-id#operation-overview)
[Mandrill](https://mailchimp.com/developer/transactional/api/metadata/)


modules/email/src/controllers/queue.controller.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  static async getInstance(grpcSdk?: ConduitGrpcSdk): Promise<QueueController> {
+    if (QueueController._instance) {
+      return QueueController._instance;
+    }
+    if (!grpcSdk) {
+      throw new Error('No grpcSdk instance provided!');
+    }
+    const instance = new QueueController(grpcSdk);
+    await instance.initializeQueues();
```
kkopanidis, line 35: initialization should occur on constructor, the get instance function is only responsible for creating the instance of the controller. No other operation should  be done here

kkopanidis: once you remove this, you must also remove the async and remove the await QueueController.getInstance() occurrences from the code

kkopanidis (changes requested):

modules/email/src/controllers/queue.controller.ts
krigol14: Done, thanks for the feedback. 

Quick question: Should we control the status polling from config (enabled/interval)? Also, the backoff strategy I've implemented is extremely basic. Should I keep it, remove it, or implement a more elaborate one?

kkopanidis: 1. I don't think so, it should rely only in the email storage being enabled since without it it won't have a sense to poll for anything.
2. The backoff strategy seems good enough for what we require I would say

kkopanidis (approved):

PR #1461, opened 2026-04-21
feat(database): stale view cleanup

ChrisPdgn:
## Stale Authorization View Cleanup

When row-level authorization is enabled on a schema, Conduit creates a MongoDB view per `(user, operation, resource type)` combination. With many users these views accumulate and can cause memory pressure on the database. The view cleanup feature automatically removes views that haven't been accessed in a configurable amount of time.

### How it works

1. Each time a view is resolved for an authorized query, its `lastAccessedAt` timestamp is updated in the `Views` collection (throttled to at most once per hour per view to avoid write overhead).
2. A BullMQ repeatable job is stored in Redis. Every database instance registers a worker against the same queue, but **each scheduled run is taken by a single worker**—you do not get one full cleanup pass per instance.
3. The worker deletes views whose `lastAccessedAt` is older than the configured threshold (up to `batchSize` per run). Deletions go through the database gRPC service, which may be handled by any instance in a load-balanced setup.
4. If a view is deleted while a query is in flight, the system catches the error, recreates the view, and retries the query transparently.
5. In multi-instance deployments, successful view removal is propagated to other instances via the `database:delete:view` Redis bus event so their in-memory caches stay consistent with MongoDB or SQL.

### Configuration

The feature is **disabled by default**. Enable and configure it via the database module's config:

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `viewCleanup.enabled` | Boolean | `false` | Enable the cleanup job |
| `viewCleanup.stalenessThresholdMs` | Number (ms) | `604800000` (7 days) | How long a view must be unused before it is deleted |
| `viewCleanup.cronPattern` | String | `""` | Cron expression for scheduling (e.g. `"0 3 * * *"` for 3 AM daily). Must be set for the cleanup job to run when enabled |
| `viewCleanup.cronTimezone` | String | `"UTC"` | IANA timezone for the cron pattern (e.g. `"America/New_York"`) |
| `viewCleanup.batchSize` | Number | `100` | Maximum number of views deleted per run |

### Scheduling recommendations

Use `cronPattern` to run cleanup during low-traffic hours so users are less likely to notice the one-time recreation latency when their view is accessed again after deletion. The cron schedule is evaluated in `cronTimezone` (IANA name, default `UTC`).

If you enable cleanup but leave `cronPattern` empty, the server logs a warning and does not schedule the job — set a non-empty cron expression to run cleanup.

Set `cronTimezone` to your region when you want local wall-clock times (for example `Europe/Athens` or `America/New_York`).

```
"0 3 * * *"   → every day at 3 AM
"0 3 * * 0"   → every Sunday at 3 AM
"0 2 * * 1-5" → weekdays at 2 AM
```

### Notes

- Views that existed before this feature was introduced (no `lastAccessedAt` field) fall back to `updatedAt` for the staleness check.
- Deleted views are recreated on demand the next time a user makes an authorized request — there is no data loss, only a small latency cost on the first request after recreation.
- On **MongoDB**, if another instance already dropped the view’s backing collection, a concurrent `deleteView` may see MongoDB’s `ns not found` error; that case is treated as already-deleted so cleanup can finish without failing the job. SQL adapters use `DROP VIEW IF EXISTS` for the same reason.
- All instances must use the **same Redis** (and thus the same BullMQ queue) for cleanup scheduling and the bus to behave as described above.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**


modules/database/src/adapters/DatabaseAdapter.ts
```diff
@@ -29,6 +29,8 @@ export abstract class DatabaseAdapter<T extends Schema> {
   protected grpcSdk: ConduitGrpcSdk;
   private legacyDeployment = false; // unprefixed declared schema collection
   private readonly _systemSchemas: Set<string> = new Set();
+  private viewAccessTimestamps: Map<string, number> = new Map();
+  private static readonly ACCESS_UPDATE_INTERVAL_MS = 60 * 60 * 1000; // 1 hour
```
kkopanidis: shouldn't this be configurable?

modules/database/src/config/config.ts
```diff
@@ -0,0 +1,30 @@
...
+      doc: 'Time in milliseconds after which an unused view is considered stale',
+      format: 'Number',
+      default: 7 * 24 * 60 * 60 * 1000,
+    },
+    cronPattern: {
+      doc: 'Cron pattern for scheduling cleanup. Required when view cleanup is enabled. Example: "0 3 * * *" for daily at 3 AM',
+      format: 'String',
+      default: '',
+    },
+    cronTimezone: {
```
kkopanidis: timezone will only resolve through the TZ env naturally. We don't want to have any skew between TZ on the machine and the app. On the UI we can add a picker for user visibility

modules/database/src/config/config.ts
```diff
@@ -0,0 +1,30 @@
...
+      doc: 'IANA timezone for the cron pattern (e.g. "America/New_York")',
+      format: 'String',
+      default: 'UTC',
+    },
+    batchSize: {
```
kkopanidis: Why do we care about batching here?

modules/database/src/jobs/cleanupStaleViews.ts
```diff
@@ -0,0 +1,57 @@
...
+  const { stalenessThresholdMs, batchSize } = job.data;
+  const threshold = new Date(Date.now() - stalenessThresholdMs);
+
+  const staleQuery: Indexable = {
+    $or: [
+      { lastAccessedAt: { $lt: threshold } },
+      { lastAccessedAt: { $exists: false }, updatedAt: { $lt: threshold } },
+    ],
+  };
+  const staleViews = (await grpcSdk.database!.findMany(
```
kkopanidis: use delete many, no reason to delete one by one

kkopanidis (changes requested):

modules/database/src/jobs/cleanupStaleViews.ts
ChrisPdgn: Deletion is handled by the deleteView rpc, we can't just delete the view docs like this. Do I miss something?

modules/database/src/config/config.ts
ChrisPdgn: In case the db has a million views?🥹

kkopanidis: It's a delete many op, I don't think there's an issue with that. But OK at least can we have a way to not do batches by default? For example set size 0 to mean "all"

modules/database/src/jobs/cleanupStaleViews.ts
kkopanidis: @ChrisPdgn yes, we shouldn't need an RPC here, since we're talking with ourselves. Same as what is happening in Authorization I'd say. We use RPC for other modules but not for self. I get the point with the findMany though, it is needed to happen first yes. But then let's create a bulk delete utility for the view deletion otherwise it will take a lot of time and resources to complete

ChrisPdgn:
@kkopanidis feature has been tested, with multiple database instances as well

PR #1463, opened 2026-04-30
fix(authentication): update logged in users metric tracking

it2022031:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**
This PR improves the accuracy of the `logged_in_users_total` authentication metric.

Changes include:
- Initialize `logged_in_users_total` based on valid refresh tokens.
- Update the metric when valid login/session tokens are issued.
- Decrement the metric when user tokens are deleted, such as during password reset or user deactivation.
- Recalculate the metric after expired token cleanup.
- Update the metric help text to reflect active token tracking.

### DIFF
--- a/modules/authentication/src/Authentication.ts
+++ b/modules/authentication/src/Authentication.ts
@@ -222,11 +222,8 @@ export default class Authentication extends ManagedModule<Config> {
     this.grpcSdk.unmonitorModule('sms');
     this.grpcSdk.unmonitorModule('router');
   }
-
   async initializeMetrics() {
-    // @improve-metrics
-    // TODO: This should initialize 'logged_in_users_total' based on valid tokens
-    // Gauge value should also decrease on latest token expiry.
+    await AuthUtils.reconcileLoggedInUsersMetric();
   }
 
   // produces login credentials for a user without them having to login
@@ -256,6 +253,7 @@ export default class Authentication extends ManagedModule<Config> {
         clientId,
         config,
       });
+      ConduitGrpcSdk.Metrics?.increment('logged_in_users_total');
       return callback(null, {
         accessToken: tokens.accessToken,
         refreshToken: tokens.refreshToken ?? undefined,
@@ -458,6 +456,7 @@ export default class Authentication extends ManagedModule<Config> {
       config,
       isRefresh: false,
     });
+    ConduitGrpcSdk.Metrics?.increment('logged_in_users_total');
     return callback(null, {
       accessToken: tokens.accessToken,
       refreshToken: tokens.refreshToken ?? undefined,
@@ -665,7 +664,11 @@ export default class Authentication extends ManagedModule<Config> {
     await User.getInstance().findByIdAndUpdate(user._id, { active });
     callback(null, { message: 'ok' });
     if (!active) {
-      TokenProvider.getInstance().deleteUserTokens({ user: id });
+      await TokenProvider.getInstance().deleteUserTokens({
+        user: id,
+      });
+
+      await AuthUtils.reconcileLoggedInUsersMetric();
     }
   }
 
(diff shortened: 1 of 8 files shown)


modules/authentication/src/Authentication.ts
```diff
@@ -222,11 +222,14 @@ export default class Authentication extends ManagedModule<Config> {
     this.grpcSdk.unmonitorModule('sms');
     this.grpcSdk.unmonitorModule('router');
   }
-
   async initializeMetrics() {
-    // @improve-metrics
-    // TODO: This should initialize 'logged_in_users_total' based on valid tokens
-    // Gauge value should also decrease on latest token expiry.
+    const validTokens = await models.RefreshToken.getInstance(this.database).findMany({
```
ChrisPdgn: I suggest to make a util function called "reconcileLoggedInUsersMetric" in the utils folder that can be called from anywhere (including in here)
It should find the count value and set it to the metrics.
Also important: you should do the findMany() query on the AccessToken schema not the RefreshToken one

modules/authentication/src/Authentication.ts
```diff
@@ -665,7 +670,13 @@ export default class Authentication extends ManagedModule<Config> {
     await User.getInstance().findByIdAndUpdate(user._id, { active });
     callback(null, { message: 'ok' });
     if (!active) {
-      TokenProvider.getInstance().deleteUserTokens({ user: id });
+      const [deletedAccessTokens] = await TokenProvider.getInstance().deleteUserTokens({
+        user: id,
+      });
+
+      for (let i = 0; i < deletedAccessTokens.deletedCount; i++) {
```
ChrisPdgn: Instead of doing this for loop, you can call the reconciliation function I mentioned above

modules/authentication/src/Authentication.ts
```diff
@@ -840,12 +851,15 @@ export default class Authentication extends ManagedModule<Config> {
...
-      models.AccessToken.getInstance()
-        .deleteMany({ expiresOn: { $lte: new Date() } })
-        .catch(e => {
-          ConduitGrpcSdk.Logger.error(`Token cleanup failed: ${e.message}`);
+    const cleanup = async () => {
+      try {
+        await models.AccessToken.getInstance().deleteMany({
+          expiresOn: { $lte: new Date() },
         });
+        await this.initializeMetrics();
```
ChrisPdgn: call in here the reconciliation function

modules/authentication/src/handlers/local.ts
```diff
@@ -512,10 +513,14 @@ export class LocalHandlers implements IAuthenticationStrategy {
     await User.getInstance().findByIdAndUpdate(user._id, { hashedPassword });
     await Token.getInstance().deleteOne(passwordResetTokenDoc);
 
-    await TokenProvider.getInstance().deleteUserTokens({
+    const [deletedAccessTokens] = await TokenProvider.getInstance().deleteUserTokens({
       user: user._id,
     });
 
+    for (let i = 0; i < deletedAccessTokens.deletedCount; i++) {
```
ChrisPdgn: same in here, call the reconciliation function

ChrisPdgn (changes requested):

modules/authentication/src/Authentication.ts
it2022031: Done. I moved the metric recalculation logic into AuthUtils.reconcileLoggedInUsersMetric() so it can be reused from multiple places.

The function now counts valid AccessToken records instead of RefreshToken records and sets logged_in_users_total based on that count.

modules/authentication/src/Authentication.ts
it2022031: Done. Replaced the manual decrement loop with a call to AuthUtils.reconcileLoggedInUsersMetric() after token deletion.

modules/authentication/src/Authentication.ts
it2022031: Done.

modules/authentication/src/handlers/local.ts
it2022031: Done.

modules/authentication/src/utils/index.ts
```diff
@@ -22,6 +22,14 @@ export namespace AuthUtils {
     return crypto.randomBytes(size).toString('base64');
   }
 
+  export async function reconcileLoggedInUsersMetric() {
+    const validTokens = await AccessToken.getInstance().findMany({
```
ChrisPdgn: You should use countDocuments() because you don't want to retrieve the actual docs but rather the count of them

ChrisPdgn (changes requested):

modules/authentication/src/utils/index.ts
it2022031: I replaced findMany() with countDocuments()

ChrisPdgn (approved):

ioanniskemerlis:
@it2022031 , @ChrisPdgn 
Screenshot 2026-05-04 at 4 12 44 PM

it works , I connected it to both prometheus & grafana.

But we can make it better if you would like. Right now , we dont have "real" statistics in case of a user re-logging and getting a new token.

As you can see in the picture, we get spikes and these were caused by incrementing by 1, but not decrementing when the previous token gets deleted. Then reconcile run and sets the count to the correct value.

What we could do is:
increment only inside "createUserTokens" since all methods use this path, and wherever we delete old tokens, we get the count of the deleted tokens, and decrement by that number

PR #1464, opened 2026-05-05
fix(database): cannot overwrite view bug

ChrisPdgn:
## Problem

Concurrent calls to `createView` with the same `viewName` caused intermittent failures:

```text
[ERROR] /database.DatabaseProvider/createView INTERNAL: Cannot overwrite … model once compiled.
```

This is a **check-then-act race**. Two or more callers both see `this.views[viewName]` as unset, then both try to register the same Mongoose model and create the same MongoDB view. The second caller fails because:

- `mongoose.model(name, schema)` throws **`OverwriteModelError`** if the name is already registered.
- `createCollection({ viewOn })` fails with **`NamespaceExists`** if the view already exists in MongoDB.

Typical triggers:

- Concurrent `CreateResourceAccessList` calls that resolve to the same hashed `viewName`.
- Multiple database instances handling `database:create:view` bus events simultaneously.
- `recoverViewsFromDatabase` running many `createView` calls in parallel.

The original code also had a second, subtler race in metadata insertion: a check-then-create pattern (`findOne` -> `create`) that could produce duplicate `Views` documents.

## Solution

The fix introduces three complementary layers of concurrency protection and separates the tightly coupled creation logic into distinct steps.

### 1. In-process promise coalescing (`pendingViewCreations`)

A `Map<string, Promise<void>>` tracks in-flight work per `viewName`. When a second caller within the same Node.js process requests the same view while creation is already running, it `await`s the existing promise instead of starting duplicate work. After the promise resolves, the caller re-checks local state before proceeding; it does not blindly assume the first caller succeeded.

### 2. Cross-instance distributed locking via Redlock

For multi-replica deployments (multiple database module instances sharing the same MongoDB and Redis), a Redlock-based distributed lock is acquired per view name.

Two methods were added to `StateManager`:

- **`tryAcquireLock(resource, ttl)`**: attempts to acquire the lock exactly once (`retryCount: 0`). Returns the lock on success, or `null` on contention. Only swallows true lock contention; Redis/infrastructure failures are re-thrown.
- **`withLock(resource, ttl, fn)`**: acquires the lock, executes `fn`, and releases in a `finally` block.

The `isLockContention` helper inspects Redlock's `ExecutionError.attempts` to distinguish true lock contention (all vote-against errors are `ResourceLockedError`) from infrastructure failures (for example, Redis unavailable), ensuring real errors surface instead of being treated as normal contention.

The `runCreateViewLocked` flow is:

1. **Try to be the writer**: call `tryAcquireLock`. If acquired, run the creation logic and release.
2. **If contended**: don't queue behind the lock. Instead, poll for readiness.
3. **If readiness times out**: fall back to a blocking `withLock` call that waits for the lock and then creates the view.

### 3. Readiness polling for lock losers (`waitForViewReadiness`)

Instead of every lock loser serially queuing to acquire the same lock, they poll for the view's existence in parallel, checking every 50ms for up to 5 seconds:

1. Is the view already in the local `this.views` cache with a matching query?
2. Does the `Views` metadata document exist in MongoDB (read from primary) with the correct query?
3. Does the physical MongoDB view namespace exist?
4. Does the MongoDB view definition (`viewOn` + `pipeline`) match what we expect?

If all checks pass, the lock loser registers the Mongoose model locally via `buildViewModel` and returns without acquiring the lock. If something doesn't match, such as a query mismatch or a namespace that exists but is not a view, a descriptive `GrpcError` is thrown rather than silently producing corrupt state.

### 4. Separation of creation concerns

The old code mixed Mongoose model registration with MongoDB view creation in one path. The new code separates this into focused private methods:

- `buildViewModel`: registers the Mongoose model in memory (`this.views[viewName]`) and clears stale models first.
- `createPhysicalMongoView`: calls `db.createCollection(viewName, { viewOn, pipeline })`.
- `upsertViewMetadata`: uses `updateOne` with `upsert: true`, replacing the old race-prone check-then-create.
- `createViewLocked`: orchestrates the full locked creation path with state checks.

## Summary of changes

- Same-process callers now coalesce on one in-flight promise per `viewName`.
- Cross-instance callers coordinate through a Redis-backed Redlock per view name.
- Lock losers verify readiness in parallel before falling back to blocking lock acquisition.
- MongoDB view creation, local Mongoose model registration, and metadata persistence are handled separately.
- View metadata is upserted instead of inserted through a check-then-create pattern.
- Redlock contention is distinguished from Redis/infrastructure failure.

## Testing

Verified with two database instances deployed concurrently. The `OverwriteModelError` no longer occurs under the previously failing scenario.

---

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

- [ ] When resolving a specific issue, it's referenced in the PR's description (e.g. `fix #xxx`, where "xxx" is the issue number)


ioanniskemerlis (approved):

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -101,25 +266,108 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
...
-    if (isNil(foundView)) {
-      await this.models['Views'].create({
-        name: viewName,
-        originalSchema: modelName,
-        joinedSchemas: [...new Set(joinedSchemas.concat(modelName))],
-        query,
-        lastAccessedAt: new Date(),
-      });
+
+    if (metadata && !isEqual(metadata.query, query)) {
```
kkopanidis, line 448: On query mismatch we should delete and recreate, not throw. After an upgrade with a changed authz pipeline the existing Views metadata will have the old query and this will 500 permanently until someone cleans up manually. We already handle this for the in-memory case right above.

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -71,15 +78,176 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
...
+      if (this.views[viewName] && isEqual(this.views[viewName]?.viewQuery, query)) {
+        return true;
+      }
+
+      const metadata = await this.models['Views'].findOne(
+        { name: viewName },
+        { readPreference: 'primary' },
+      );
+
+      if (metadata && !isEqual(metadata.query, query)) {
```
kkopanidis: Same issue here — lock losers hit this during polling and get a hard failure instead of self-healing.

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -101,25 +266,108 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
...
+    return viewModel;
+  }
 
-    const foundView = await this.models['Views'].findOne(
+  private async createPhysicalMongoView(
+    viewName: string,
+    viewOn: string,
+    pipeline: object[],
+  ): Promise<void> {
+    await this.mongoose.connection.db!.createCollection(viewName, {
```
kkopanidis: Catch `NamespaceExists` here and reconcile through the existing definition check path. Rolling deploys can still hit this race between the namespace check and `createCollection`.

modules/database/src/adapters/mongoose-adapter/index.ts
```diff
@@ -71,15 +78,176 @@ export class MongooseAdapter extends DatabaseAdapter<MongooseSchema> {
...
+      modelName,
+      viewName,
+      joinedSchemas,
+      query,
+    );
+    if (ready) {
+      return;
+    }
+
+    return state.withLock(resource, VIEW_LOCK_TTL_MS, () =>
```
kkopanidis: This gives up after Redlock's default retries (~2s) on top of the 5s poll. If the holder is legitimately slow, callers still get a 500. Consider extending retries for view creation specifically, or continuing readiness polling instead of failing fast.

kkopanidis (changes requested):
The concurrency approach is correct and addresses the production issue. A few things before merge though:

1. Metadata query mismatch throws INTERNAL — on upgrade this will permanently brick authz views with stale metadata until manual cleanup. Restore the old self-heal behavior (`deleteView` + recreate), same as we already do for the in-memory query mismatch case.
2. `createPhysicalMongoView` needs to catch `NamespaceExists` and reconcile. Rolling deploys can still race here and surface the same error we're fixing.
3. The lock fallback isn't really blocking — 5s poll + ~2s of Redlock retries then we throw. Slow holders will still produce 500s even though the view may appear seconds later. Either extend the retry budget for view creation or keep polling.
4. Would like a test here — two adapter instances, shared mongo/redis, concurrent `createView` for the same `viewName`. No throw, one physical view, all instances registered.

modules/database/src/adapters/mongoose-adapter/index.ts
ChrisPdgn: > On query mismatch we should delete and recreate, not throw. After an upgrade with a changed authz pipeline the existing Views metadata will have the old query and this will 500 permanently until someone cleans up manually. We already handle this for the in-memory case right above.

Addressed in `cd85a4504`. `createViewLocked` now calls `reconcileStaleView(viewName)` when metadata query (or local in-memory query / incompatible physical definition) does not match the requested query. That helper drops the physical MongoDB view only when `listCollections` confirms it is a view, deletes the `Views` metadata row, and clears the local Mongoose model — then creation proceeds normally via `ensurePhysicalMongoView` + `upsertViewMetadata`.

This restores the self-heal behavior without relying on `deleteView`, which only dropped the physical namespace when the view was already registered locally.

modules/database/src/adapters/mongoose-adapter/index.ts
ChrisPdgn: > Same issue here — lock losers hit this during polling and get a hard failure instead of self-healing.

Addressed in `cd85a4504`. Polling no longer throws on stale state. `assessViewReadinessOnce` returns `needs_reconcile` for metadata query mismatch or incompatible physical definition, and `waitForViewReadiness` exits early with `false` so the caller falls through to the locked writer path.

Lock losers do not call `reconcileStaleView` while polling — only the lock holder performs destructive cleanup. If the view is already ready with a matching definition, losers still call `buildViewModel` and return immediately.

modules/database/src/adapters/mongoose-adapter/index.ts
ChrisPdgn: > Catch `NamespaceExists` here and reconcile through the existing definition check path. Rolling deploys can still hit this race between the namespace check and `createCollection`.

Addressed in `cd85a4504`. `createPhysicalMongoView` was replaced by `ensurePhysicalMongoView`, which catches `NamespaceExists` (code 48 / `codeName`), re-reads the physical definition, and accepts it when `viewOn` + `pipeline` match. On mismatch it runs `reconcileStaleView` under the existing lock and retries `createCollection` once.

modules/database/src/adapters/mongoose-adapter/index.ts
ChrisPdgn: > This gives up after Redlock's default retries (~2s) on top of the 5s poll. If the holder is legitimately slow, callers still get a 500. Consider extending retries for view creation specifically, or continuing readiness polling instead of failing fast.

Addressed in `cd85a4504`. The `state.withLock` fallback was replaced by `waitForViewOrCreateWhenUnlocked`, which keeps alternating readiness checks and `tryAcquireLock` attempts for up to `VIEW_LOCK_TTL_MS` (60s) instead of Redlock's default ~2s retry budget.

Flow is unchanged for the fast path: initial 5s `waitForViewReadiness` poll, then the extended loop only if needed. Timeout now throws a descriptive `GrpcError` with the view name and deadline.

ChrisPdgn:
Thanks for the review — all four points are addressed in `cd85a4504` (`fix: address review`):

1. **Metadata query mismatch** — `createViewLocked` now calls `reconcileStaleView` for stale metadata / incompatible physical definitions (not just in-memory mismatch). Drops the physical view only when confirmed to be a MongoDB view, clears metadata + local model, then recreates.

2. **`NamespaceExists` race** — `ensurePhysicalMongoView` catches the error, validates the existing definition, reconciles on mismatch, and retries creation.

3. **Lock fallback timeout** — replaced `state.withLock` fallback with `waitForViewOrCreateWhenUnlocked`: alternates readiness checks + `tryAcquireLock` for up to 60s (`VIEW_LOCK_TTL_MS`) after the initial 5s poll.

4. **Integration test** — added `modules/database/src/__tests__/integration/mongoose-adapter.createView.concurrent.test.ts` (gated on `DB_CONN_URI` + `REDIS_HOST` + `REDIS_PORT`). Two `MongooseAdapter` instances, shared Mongo/Redis, concurrent `createView` for the same `viewName` — asserts no throw, one physical view, one metadata row, both adapters registered. Run via `pnpm test:integration` in `modules/database`.

Lock-loser polling nuance: stale state returns `needs_reconcile` during poll (no throw, no delete); only the lock holder runs `reconcileStaleView`.

PR #1468, opened 2026-05-07
fix(authentication): reconcile logged in users metric

it2022031:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

**Other information:**

This PR improves the accuracy of the `logged_in_users_total` authentication metric.

Previously, the metric could become inconsistent because expired sessions were not always removed correctly when tokens expired naturally over time. As a result, `logged_in_users_total` could temporarily report incorrect values.

Additionally, the metric was effectively tracking active tokens instead of unique logged-in users, meaning the same user could be counted multiple times across different sessions.

This PR centralizes logged-in user metric updates in `TokenProvider` and improves cleanup/reconciliation logic so the metric reflects unique active logged-in users more reliably across authentication flows, token refreshes, session invalidation, and expired token cleanup.

No breaking changes are introduced.


kkopanidis (changes requested):
Unnecessary duplication present. The "provideUserTokens" function can be utilized to call the "addLoggedInUser" and "reconcileLoggedInUserMetric", instead of adding it everywhere in the code

modules/authentication/src/handlers/tokenProvider.ts
```diff
@@ -140,7 +146,19 @@ export class TokenProvider {
 
     return Promise.all([promise1, promise2]);
   }
+  private async trackLoggedInUserMetric(
```
ChrisPdgn, line 153: You should add try/catch inside this method because we want to make sure this doesn't throw an error and disrupt/break actual authentication. 
When you catch the error you can log a warning like this
```
ConduitGrpcSdk.Logger.warn(`Logged-in user metric tracking failed`);
ConduitGrpcSdk.Logger.warn(err);
```
Also, when called, I think we shouldn't await this

modules/authentication/src/handlers/common.ts
```diff
@@ -77,14 +79,18 @@ export class CommonHandlers implements IAuthenticationStrategy {
...
     await TokenProvider.getInstance().logOutClientOperations(
       this.grpcSdk,
       clientConfig,
       authToken!,
       clientId,
       user._id,
     );
+
+    if (!clientConfig.multipleUserSessions) {
+      await AuthUtils.removeLoggedInUser(user._id);
```
ChrisPdgn: Create a wrapper that will call these functions in a try/catch and don't await
Adjust other similar places

modules/authentication/src/handlers/magicLink.ts
```diff
@@ -128,7 +128,7 @@ export class MagicLinkHandlers implements IAuthenticationStrategy {
     const redirectUri =
       AuthUtils.validateRedirectUri(data.redirectUri) ??
       config.magic_link.redirect_uri.replace(/\/$/, '');
-    return TokenProvider.getInstance().provideUserTokens(
+    return TokenProvider.getInstance()!.provideUserTokens(
```
ChrisPdgn: why did we add this '!'?

ChrisPdgn (changes requested):

modules/authentication/src/handlers/tokenProvider.ts
it2022031: Added try/catch around logged-in user metric tracking and changed the calls to fire-and-forget so metric failures do not interrupt authentication.

modules/authentication/src/handlers/common.ts
it2022031: Added a wrapper for logged-out user metric cleanup and updated the related call sites to use it without awaiting metric operations.

modules/authentication/src/handlers/magicLink.ts
it2022031: Removed the unnecessary non-null assertion.

ChrisPdgn (approved):

kkopanidis (approved):

PR #1486, opened 2026-06-18
fix(hermes): support Socket.IO handshake auth

kkopanidis:
## Summary
- Bridge Socket.IO `handshake.auth` token fields into Hermes socket middleware headers so browser websocket-first clients can authenticate without polling `extraHeaders`.
- Preserve existing transport `Authorization` headers and keep client ID/secret websocket-first support out of scope.
- Add focused Hermes unit coverage with a test-only TypeScript build output.

## Test plan
- `cd libraries/hermes && npm test`


libraries/hermes/src/Socket/buildSocketMiddlewareParams.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+
+  return {
+    headers,
+    cookies: {},
+    context: {
+      ...((socket.request as { conduit?: Record<string, unknown> }).conduit ?? {}),
+      ...(socket.data ?? {}),
+    },
```
ChrisPdgn, line 37: **Bug — context reference severed; event handlers lose middleware results**

The spread copy here creates a new object disconnected from `socket.request.conduit`. In the old code, `context.context` was assigned `socket.request.conduit` **by reference**, so when `checkMiddlewares` mutated it (e.g. the auth middleware merging `{ user: {...} }`), `socket.request.conduit` was updated in place.

Now the mutations only affect this detached copy. `socket.data` is correctly set afterward, but the three event handlers (`connect`, `onAny`, `disconnect`) in `Socket.ts` all still read from `socket.request.conduit`:

```ts
context: socket.request.conduit, // lines 147, 164, 182
```

Since `conduitRequestMiddleware` and `ClientValidator` are both registered with `socketMiddleware: true`, `socket.request.conduit` is always truthy (`{ clientId: '...' }`), so the old `||` fallback never kicked in — the reference was always kept.

**Result:** after this PR, `connect`/`onAny`/`disconnect` handlers receive `{ clientId }` but **no user data** — breaking authenticated socket connections.

**Suggested fix:** either write the enriched context back after `checkMiddlewares`:
```ts
(socket.request as any).conduit = context.context;
```
or update the event handlers to read from `socket.data` instead (which is already correctly assigned).

libraries/hermes/src/Socket/buildSocketMiddlewareParams.ts
```diff
@@ -0,0 +1,34 @@
...
+export function buildSocketMiddlewareParams(socket: Socket): SocketMiddlewareParams {
+  const headers = { ...socket.request.headers };
+  const auth = socket.handshake.auth ?? {};
+
+  const hasAuthHeader =
+    headers.authorization !== undefined || headers.Authorization !== undefined;
+
+  if (!hasAuthHeader) {
+    const rawToken = auth.token ?? auth.accessToken;
+    if (rawToken) {
```
ChrisPdgn, line 24: **Double-`Bearer` risk**

If a client passes `auth.token = "Bearer eyJ..."` (already prefixed), this produces `"Bearer Bearer eyJ..."`. Consider stripping an existing prefix:

```ts
if (rawToken) {
  const tokenStr = String(rawToken);
  headers.authorization = tokenStr.startsWith('Bearer ')
    ? tokenStr
    : `Bearer ${tokenStr}`;
}
```

ChrisPdgn (changes requested):
The handshake.auth bridging logic is a solid addition, but the refactor severs a reference that the existing event handlers depend on, breaking authenticated socket connections.

libraries/hermes/src/Socket/buildSocketMiddlewareParams.ts
kkopanidis: This is addressed on the current PR by having the socket handlers read from `socket.data` instead of `socket.request.conduit`.

The namespace middleware mutates `context.context`, then assigns it to `socket.data`, and `connect` / `onAny` / `disconnect` now pass `context: socket.data` to module handlers. That keeps the middleware-enriched auth context available without relying on the old `socket.request.conduit` object reference.

ChrisPdgn (approved):

PR #1487, opened 2026-06-18
fix(hermes): support Socket.IO handshake auth

kkopanidis:
## Summary
- Bridge Socket.IO `handshake.auth` token fields into Hermes socket middleware headers so browser websocket-first clients can authenticate without polling `extraHeaders`.
- Preserve existing transport `Authorization` headers, ignore socket cookies, and pass `socket.data` context to socket handlers on main.
- Add focused Hermes unit coverage with a test-only TypeScript build output.

## Test plan
- `cd libraries/hermes && pnpm test`


ChrisPdgn:
## Review Notes

### 1. Token precedence is silently lossy

In `buildSocketMiddlewareParams.ts`, when a client provides both `auth.token` (or `auth.accessToken`) **and** `auth.authorization`, the second `if` block silently overwrites the first:

```ts
const rawToken = auth.token ?? auth.accessToken;
if (rawToken && !hasAuthorizationHeader) {
  headers.authorization = `Bearer ${rawToken}`;
}
if (typeof auth.authorization === 'string' && !hasAuthorizationHeader) {
  headers.authorization = auth.authorization;
}
```

If a client sends `{ token: 'abc', authorization: 'Basic xyz' }`, the `Basic` scheme wins with no warning. Consider using an `else if` to make the priority chain explicit, or at least document the intended precedence.

### 2. Behavioral change: event handlers now receive post-middleware context

Previously, `connect`, arbitrary events, and `disconnect` handlers all read from `socket.request.conduit` (the raw pre-middleware context). Now they read from `socket.data`, which contains the **post-middleware enriched** context. This is likely the desired behavior, but it's a semantic breaking change for any socket handler that relied on receiving only the raw context without middleware augmentations. Worth calling out in release notes.

ChrisPdgn (approved):

PR #1488, opened 2026-06-18
feat(routing): raw headers and opt-in raw body for webhook verification

ChrisPdgn:
## Summary
- Forward sanitized `rawHeaders` and opt-in `rawBody` through Hermes/gRPC so webhook routes can verify signatures.
- Add `captureRawBody` to routing types and proto definitions, including the follow-up proto fix.
- Keep stripped/security-sensitive headers out of forwarded `rawHeaders`.

## Test plan
- [x] Manual webhook signature verification with `captureRawBody: true`
- [x] Confirm stripped credentials are omitted from `rawHeaders`


kkopanidis:
@ChrisPdgn — follow-up commit pushed addressing review feedback from the three-horsemen review:

- **`captureRawHeaders` opt-in flag** (proto field 15, `RouteBuilder` helpers) — raw headers only forwarded over gRPC when enabled
- **Form-encoded webhook support** — shared `captureRequestRawBody` verify on both `express.json` and `express.urlencoded`
- **Per-route gating** — strips `rawBody`/`rawHeaders` when flags are off; clears `req.rawBody` early to release buffer reference
- **`sanitizeRawHeaders` guard** — safe on missing `rawHeaders` (mocks/tests)
- **GrpcConverter** — skips serializing empty `rawHeaders` arrays

Webhook routes should use both:
```ts
.captureRawBody(true)
.captureRawHeaders(true)
```

Ready for your review when you have a moment.

kkopanidis (approved):

PR #1507, opened 2026-06-21
feat(functions): implement cron scheduling with BullMQ

kkopanidis:
## Problem

`functionType: 'cron'` was accepted by the admin API and UI but never ran on a schedule. The runtime fell through to the event-bus path and subscribed to a channel named after the cron expression, so scheduled functions silently never executed.

## Summary

- Add BullMQ `functions-cron-queue` with idempotent `syncCronJobs()` on every `refreshRoutes()`
- Stable repeatable job IDs (`cron-{functionId}`); orphan cleanup on delete/type change
- Canonical `inputs.cronPattern` with legacy `inputs.event` alias; admin validation on upload/patch
- Cron invocations reuse the VM sandbox, `FunctionExecutions`, and existing metrics
- Migration copies legacy `inputs.event` → `inputs.cronPattern` for existing cron docs

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Test plan

- [ ] Create a cron function via Admin API (`*/5 * * * *`); confirm `FunctionExecutions` records appear within ~5 minutes
- [ ] Update the cron pattern; confirm cadence changes without duplicate Redis repeatable keys
- [ ] Delete the function; confirm executions stop after refresh
- [ ] Verify `functionType: 'event'` still subscribes to the bus (no regression)
- [ ] Set functions module `active: false`; confirm queue drains and ticks stop
- [ ] Rolling deploy with two replicas: single execution per tick


ChrisPdgn:
## Review notes

Items from review that we are **not** changing, and why:

- **Job-id matching** — Withdrawn. `repeatable.id === cron-{functionId}` is correct in this queue; matching is unchanged.
- **Remove-then-add** — Residual only: pattern/tz change **and** `queue.add` fails after remove. Next successful sync reschedules. Not a merge blocker.
- **Await admin refresh** — Intentional fire-and-forget. Operators inspect Redis via `GET /list/cron-jobs`.
- **Controller tests** — Out of scope for this pass. Existing `cron.utils` tests stay (updated so `event` is kept).
- **event vs cronPattern** — `cronPattern` is the schedule. We do not strip or dual-write `event`; runtime prefers `cronPattern`.
- **GitOps import** — Pre-existing import path; not part of the cron scheduler fix.

ChrisPdgn (approved):

PR #1511, opened 2026-06-21
fix(authentication): require sudo for admin team delete

kkopanidis:
## Problem

The client API already requires a sudo JWT to delete a team, but the admin API allowed `DELETE /authentication/teams/:teamId` with any valid admin session. Conduit-UI called that endpoint directly, so team deletion bypassed the sudo gate.

## Solution

- Issue admin JWTs with `sudo: true` after successful login and 2FA verification
- Attach decoded `jwtPayload` to admin request context for module handlers
- Reject admin team delete when `jwtPayload.sudo` is missing (same error as client API)
- Preserve existing cleanup: relation deletion and `parentTeam` unset on child teams

## Test plan

- [ ] Log in to Conduit-UI and delete a non-default team — should succeed with a fresh session
- [ ] Call admin `DELETE /authentication/teams/:id` with a JWT that lacks `sudo` — expect 403 `Re-login required to enter sudo mode`
- [ ] Verify child teams lose `parentTeam` reference after parent deletion
- [ ] Verify default team still cannot be deleted

### DIFF
--- a/modules/authentication/src/admin/team.ts
+++ b/modules/authentication/src/admin/team.ts
@@ -185,7 +185,7 @@ export class TeamsAdmin {
           teamId: ConduitObjectId.Required,
         },
         name: 'DeleteTeam',
-        description: 'Deletes a team',
+        description: 'Deletes a team (requires sudo access)',
       },
       new ConduitRouteReturnDefinition('DeleteTeam', 'String'),
       this.deleteTeam.bind(this),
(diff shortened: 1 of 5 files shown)


kkopanidis:
Closing: sudo applies to Client API end-user actions (e.g. DELETE /authentication/teams with user bearer token), not Admin API. Conduit-UI team delete uses admin routes and should not require sudo.

PR #1512, opened 2026-06-21
fix(chat): invitation email hook paths and login return flow

kkopanidis:
## Problem

Chat invitation emails link to `/hook/chat/invitations/{accept|decline}/{token}`, but the hook handler required an authenticated user in request context and returned a plain string response. Users clicking email links while logged out hit a broken flow with no redirect to login or back into the app.

## Solution

- Register the hook at `/hook/chat/invitations/:answer/:invitationToken` to match email URLs (Hermes already prefixes module hooks, but the route is now explicit).
- Add `authMiddleware?` so the hook works for anonymous email-link visits.
- Look up invitations by token only; verify the logged-in user matches the invitation receiver before mutating state.
- Redirect unauthenticated users to `explicit_room_joins.redirect.login_uri` with `redirectUri` set to the hook URL so they can return after login.
- Redirect authenticated users after accept/decline when `accept_uri` / `decline_uri` are configured (`{roomId}` placeholder supported).
- Centralize hook URL construction in `buildInvitationHookUrl` shared by email templates and the hook handler.

## Config

Set via Chat module config (`explicit_room_joins.redirect`):

| Key | Purpose |
|-----|---------|
| `login_uri` | App login page for unauthenticated email-link clicks |
| `accept_uri` | Post-accept destination (supports `{roomId}`) |
| `decline_uri` | Post-decline destination |

## Test plan

- [ ] Send a chat room invitation with email enabled
- [ ] Click accept link while logged out → redirected to `login_uri?redirectUri=<hook url>`
- [ ] Log in and return to hook URL → invitation accepted, redirected to `accept_uri`
- [ ] Click decline link while authenticated as invitee → invitation declined, redirected to `decline_uri`
- [ ] Click accept link as a different authenticated user → 403
- [ ] Click accept link when already a room member → token cleaned up, redirect/message without error


ioanniskemerlis:
## Bugs and Unintended Side Effects

### 1. Missing Membership Cache Invalidation

**Severity:** Medium — real user impact

`addUsersToRoom` in `modules/chat/src/routes/index.ts` calls `invalidateMembershipCache` after adding participants, but `processInvitationAnswer` in `modules/chat/src/routes/InvitationRoutes.ts` does not.

`verifyRoomMembership` caches participant lists for 30 seconds via `MEMBERSHIP_CACHE_TTL_MS`. After accepting an invitation through the email hook, the invitee may fail `sendMessage` or other room membership checks until the cache expires because the cached participant list still omits them.

**Fix:** Clear `chat:membership:{roomId}` in `processInvitationAnswer` after a successful accept, using the same one-liner pattern as `ChatRoutes.invalidateMembershipCache`.

---

**Severity:** Minor — not blockers


### 2. Missing Validation in Authenticated `answerInvitation`

`answerInvitation`, used by the authenticated `/chat/invitations/:answer/:id` route, lacks the `accept` / `decline` validation that exists in the hook flow.

This appears to be a pre-existing inconsistency.

ChrisPdgn (approved):

PR #1513, opened 2026-06-21
fix(chat): emit socket and bus events on admin room delete

kkopanidis:
## Problem

`DELETE /chat/rooms` (admin `deleteRooms`) bulk-deletes rooms and messages but never notifies connected clients or bus subscribers. Participants keep stale socket room membership, membership cache is not invalidated, and downstream handlers never receive `chat:deleteRoom:ChatRoom`.

Client `leaveRoom` and admin `removeUsersFromRoom` already publish bus events and push `leave-room` when a room is removed.

## Solution

- Fetch rooms before delete to capture participant lists
- Respect `auditMode`: soft-delete (`deleted: true`) vs hard `deleteMany`
- Per deleted room:
  - Invalidate `chat:membership:{roomId}` cache
  - Publish `chat:deleteRoom:ChatRoom` with `{ roomId }`
  - For each participant: `leave-room` socket push + `chat:leaveRoom:ChatRoom` bus event

## Test plan

- [ ] Admin delete single room with participants — clients receive `leave-room`, bus receives `chat:deleteRoom:ChatRoom` and per-user `chat:leaveRoom:ChatRoom`
- [ ] Admin bulk delete multiple rooms — each room's participants notified independently
- [ ] With `auditMode: true` — rooms/messages marked deleted, same notifications emitted
- [ ] With `auditMode: false` — rooms/messages hard-deleted, same notifications emitted

- [x] Bugfix

- [ ] Yes
- [x] No


modules/chat/src/admin/index.ts
```diff
@@ -169,16 +169,59 @@ export class AdminHandlers {
...
-    await ChatRoom.getInstance()
-      .deleteMany({ _id: { $in: ids } })
-      .catch((e: Error) => {
-        throw new GrpcError(status.INTERNAL, e.message);
-      });
-    await ChatMessage.getInstance()
-      .deleteMany({ room: { $in: ids } })
+
+    const rooms = await ChatRoom.getInstance()
+      .findMany({ _id: { $in: ids } })
```
ChrisPdgn: **Bug:** This query fetches all rooms matching the IDs, including ones already soft-deleted. In `auditMode` deployments, re-deleting a previously soft-deleted room will re-notify all its participants with redundant `leave-room` socket pushes and `chat:leaveRoom:ChatRoom` bus events.

`removeUsersFromRoom` and `leaveRoom` both guard against this (`if (isNil(room) || room.deleted)`).

```suggestion
      .findMany({ _id: { $in: ids }, deleted: { $ne: true } })
```

modules/chat/src/admin/index.ts
```diff
@@ -169,16 +169,59 @@ export class AdminHandlers {
...
       .catch((e: Error) => {
         throw new GrpcError(status.INTERNAL, e.message);
       });
+
+    const auditMode = ConfigController.getInstance().config.auditMode;
```
ChrisPdgn, line 182: **Suggestion:** Consider adding an early return when no rooms were found. If all provided IDs are invalid (or already soft-deleted with the filter above), the code currently still runs `deleteMany`/`updateMany` (no-ops) and decrements the metric at the end.

```typescript
if (rooms.length === 0) return 'Done';
```

modules/chat/src/admin/index.ts
```diff
@@ -169,16 +169,59 @@ export class AdminHandlers {
...
+          throw new GrpcError(status.INTERNAL, e.message);
+        });
+      await ChatMessage.getInstance()
+        .deleteMany({ room: { $in: ids } })
+        .catch((e: Error) => {
+          throw new GrpcError(status.INTERNAL, e.message);
+        });
+    }
+
+    for (const room of rooms) {
```
ChrisPdgn, line 242: **Missing audit trail:** Both `removeUsersFromRoom` (admin) and `leaveRoom` (client) create `ChatParticipantsLog` entries before removing users (`action: 'remove'` / `action: 'leave'`). This loop skips that entirely.

In `auditMode` — where the purpose is preserving an audit trail — there'll be no record of who was in the room at the time of deletion. If a room is ever un-soft-deleted, the participants log will have a gap.

Consider adding log entries (e.g. `action: 'delete'`) for each participant per room before or inside this loop.

modules/chat/src/admin/index.ts
```diff
@@ -169,16 +169,59 @@ export class AdminHandlers {
...
+        .deleteMany({ _id: { $in: ids } })
+        .catch((e: Error) => {
+          throw new GrpcError(status.INTERNAL, e.message);
+        });
+      await ChatMessage.getInstance()
+        .deleteMany({ room: { $in: ids } })
+        .catch((e: Error) => {
+          throw new GrpcError(status.INTERNAL, e.message);
+        });
+    }
```
ChrisPdgn, line 240: **Missing cleanup:** Pending `InvitationToken` documents for the deleted rooms are not cleaned up. In non-audit mode this leaves orphaned invitation documents referencing rooms that no longer exist. The `deleteRoomInvitations` admin route exists but is never called internally here.

Consider adding:
```typescript
await InvitationToken.getInstance()
  .deleteMany({ room: { $in: ids } })
  .catch((e: Error) => {
    throw new GrpcError(status.INTERNAL, e.message);
  });
```
(or `updateMany` with `{ deleted: true }` in audit mode).

modules/chat/src/admin/index.ts
```diff
@@ -169,16 +169,59 @@ export class AdminHandlers {
...
+          rooms: [roomId],
+        });
+        this.grpcSdk.bus?.publish(
+          'chat:leaveRoom:ChatRoom',
+          JSON.stringify({ roomId, user: userId }),
+        );
+      }
+    }
+
     ConduitGrpcSdk.Metrics?.decrement('chat_rooms_total');
```
ChrisPdgn: **Bug:** This decrements by 1 regardless of how many rooms were actually deleted. `createRoom` increments by 1 per room, so a bulk delete of N rooms will leave the metric permanently inflated by N−1.

Also fires unconditionally even when no rooms matched the query.

```suggestion
    ConduitGrpcSdk.Metrics?.decrement('chat_rooms_total', rooms.length);
```

ChrisPdgn (changes requested):
Good fix — the core intent is right and addresses a real notification gap. The `auditMode` branching and event emission patterns are consistent with the rest of the chat module. A few issues need addressing before merge though.

modules/chat/src/admin/index.ts
kkopanidis: Fixed in 92a42d16 — `findMany` now filters with `deleted: { $ne: true }`.

modules/chat/src/admin/index.ts
kkopanidis: Fixed in 92a42d16 — early return when `rooms.length === 0` before any DB mutations or metric decrement.

modules/chat/src/admin/index.ts
kkopanidis: Fixed in 92a42d16 — per-room `ChatParticipantsLog` entries with `action: 'remove'` (schema enum has no `delete`), merged into `participantsLog` before soft-delete in auditMode.

modules/chat/src/admin/index.ts
kkopanidis: Fixed in 92a42d16 — `InvitationToken.deleteMany({ room: { $in: roomIds } })` in both audit and non-audit paths. Hard delete only since `InvitationToken` has no `deleted` field.

modules/chat/src/admin/index.ts
kkopanidis: Addressed in aca5ed05 — `decrement('chat_rooms_total', rooms.length)`. Combined with the early return in 92a42d16 so it only fires when rooms were actually found.

ChrisPdgn (approved):

PR #1547, opened 2026-07-08
feat(auth)!: support multiple Apple OAuth client credential sets

kkopanidis:
## Summary
- Add Apple-only `clients[]` configuration for additional OAuth credential sets per app.
- Support optional `oauthClientId` on Apple init routes and resolve the selected client during callback/native authorize.
- Keep generic OAuth providers on flat `clientId` / `clientSecret` / redirect behavior.

## Test plan
- [x] `nvm use 24 && pnpm run build` in `modules/authentication`
- [ ] Configure Apple default credentials plus one additional app client via Admin API
- [ ] Start Apple OAuth with and without `oauthClientId` and verify callback/native authorize succeeds
- [ ] Confirm generic OAuth providers do not accept `oauthClientId`


ChrisPdgn:
### Summary:

Apple Sign In can now have extra client credential sets under apple.clients[] (for example a web Services ID and an iOS bundle ID), not just the one top-level client.

Each extra row needs a unique nickname (id) and a clientId. Leave privateKey / teamId / keyId empty to reuse the default key and team, or set all three for a second Apple team. Mixed is rejected on save. Empty extra redirect_uri falls back to the default. Extra clients cannot turn Apple on by themselves.

Callers pass optional oauthClientId on Apple init / initNative (our nickname, not Apple’s id). Omit it and login is unchanged. Other OAuth providers are untouched. Admin saves clients[] as objects.

BREAKING CHANGE: keyId is now required on the default Apple config to enable the provider. A config with only clientId + privateKey + teamId will stay off.

ChrisPdgn (approved):

PR #1557, opened 2026-07-20
feat(storage): add URL upload/download support for local storage provider

ChrisPdgn:
## Summary

- Add an embedded HTTP file server to the local storage provider so `getUploadUrl`, `getSignedUrl`, and `getPublicUrl` work the same way as cloud providers (presigned-URL pattern on port 3100 by default).
- Implement provider lifecycle cleanup via `dispose()` so the HTTP server is torn down and recreated on config changes (activate, deactivate, provider switch).
- Fix local provider directory creation: `store()` now creates parent dirs before writing, and `createFolder()` uses synchronous recursive `mkdirSync` instead of racy async `mkdir` callbacks.

## Test plan

- [x] `POST /storage/upload` with a new folder path succeeds without manually creating directories
- [x] `PUT` to the returned upload URL writes file bytes to disk
- [x] `GET` on the local storage URL serves the file with correct MIME type (text + image)
- [x] `GET /storage/getFileUrl/:id` returns the local storage URL for private files
- [ ] Verify HTTP server restarts cleanly when storage config changes (port/path/provider switch)
- [ ] Verify module deactivation calls `dispose()` and releases port 3100

### DIFF
--- a/modules/storage/src/Storage.ts
+++ b/modules/storage/src/Storage.ts
@@ -209,6 +209,14 @@ export default class Storage extends ManagedModule<Config> {
         config.aws.usePathStyle = false;
       }
     }
+    if (config.provider === 'local') {
+      const secret = config.local?.signingSecret?.trim() ?? '';
+      if (secret && !/^[0-9a-fA-F]{64}$/.test(secret)) {
+        throw new Error(
+          "local.signingSecret must be a 64-character hex string (32 bytes). Generate one with: node -e \"console.log(require('crypto').randomBytes(32).toString('hex'))\"",
+        );
+      }
+    }
     return config;
   }
 
--- a/modules/storage/src/Storage.ts
+++ b/modules/storage/src/Storage.ts
@@ -221,6 +221,9 @@ export default class Storage extends ManagedModule<Config> {
       this._storageAuthzResourceDispose?.();
       this._storageAuthzResourceDispose = null;
       this.disposeDeclaredPeerWatches();
+      if (this.storageProvider?.dispose) {
+        await this.storageProvider.dispose();
+      }
       this.updateHealth(HealthCheckStatus.NOT_SERVING);
     } else {
       this.registerDeclaredPeerWatches();
@@ -238,6 +241,9 @@ export default class Storage extends ManagedModule<Config> {
       } else {
         this._storageAuthzResourceDispose?.();
         this._storageAuthzResourceDispose = null;
+      }
+      if (this.storageProvider?.dispose) {
+        await this.storageProvider.dispose();
       }
       this.storageProvider = createStorageProvider(provider, {
         local,
--- a/modules/storage/src/Storage.ts
+++ b/modules/storage/src/Storage.ts
@@ -51,6 +51,7 @@ import {
 import { StorageParamAdapter } from './adapter/StorageParamAdapter.js';
 import { FileResource } from './authz/index.js';
 import { AdminFileHandlers } from './admin/adminFile.js';
+import { randomBytes } from 'node:crypto';
 import { fileURLToPath } from 'node:url';
 
 const __filename = fileURLToPath(import.meta.url);
@@ -210,11 +211,15 @@ export default class Storage extends ManagedModule<Config> {
       }
     }
     if (config.provider === 'local') {
-      const secret = config.local?.signingSecret?.trim() ?? '';
-      if (secret && !/^[0-9a-fA-F]{64}$/.test(secret)) {
+      const secret = config.local.signingSecret?.trim() ?? '';
+      if (!secret) {
+        config.local.signingSecret = randomBytes(32).toString('hex');
+      } else if (!/^[0-9a-fA-F]{64}$/.test(secret)) {
         throw new Error(
           "local.signingSecret must be a 64-character hex string (32 bytes). Generate one with: node -e \"console.log(require('crypto').randomBytes(32).toString('hex'))\"",
         );
+      } else {
+        config.local.signingSecret = secret;
       }
     }
     return config;
--- a/modules/storage/src/config/config.ts
+++ b/modules/storage/src/config/config.ts
@@ -87,6 +87,11 @@ export default {
       default: '',
       doc: 'External base URL for local storage file URLs. Leave empty to auto-derive from httpPort (http://localhost:{httpPort}). Override when the storage server is behind a reverse proxy or running in Docker.',
     },
+    signingSecret: {
+      format: 'String',
+      default: '',
+      doc: 'Hex-encoded HMAC secret for signing file URLs. If empty, a random secret is generated on startup (URLs will not survive restarts).',
+    },
   },
   suffixOnNameConflict: {
     format: 'Boolean',
--- a/modules/storage/src/config/config.ts
+++ b/modules/storage/src/config/config.ts
@@ -77,6 +77,16 @@ export default {
       format: 'String',
       default: '/var/tmp',
     },
+    httpPort: {
+      format: 'Number',
+      default: 3100,
+      doc: 'Port for the local storage HTTP file server (serves uploads/downloads)',
+    },
+    httpBaseUrl: {
+      format: 'String',
+      default: 'http://localhost:3100',
+      doc: 'Base URL for local storage file URLs (override for Docker/CI)',
+    },
   },
   suffixOnNameConflict: {
     format: 'Boolean',
--- a/modules/storage/src/config/config.ts
+++ b/modules/storage/src/config/config.ts
@@ -90,7 +90,7 @@ export default {
     signingSecret: {
       format: 'String',
       default: '',
-      doc: 'Hex-encoded HMAC secret for signing file URLs. If empty, a random secret is generated on startup (URLs will not survive restarts).',
+      doc: 'Hex-encoded HMAC secret (64 hex chars / 32 bytes) for signing file URLs. Required for URLs to survive restarts — without it, a random secret is generated on startup and any public file URLs already stored in the database (sourceUrl/url) become invalid.',
     },
   },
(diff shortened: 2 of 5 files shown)


modules/storage/src/config/config.ts
```diff
@@ -77,6 +77,16 @@ export default {
       format: 'String',
       default: '/var/tmp',
     },
+    httpPort: {
+      format: 'Number',
+      default: 3100,
+      doc: 'Port for the local storage HTTP file server (serves uploads/downloads)',
+    },
+    httpBaseUrl: {
```
ioanniskemerlis, line 85: httpPort and httpBaseUrl are independent config keys. Changing only httpPort (e.g. to 3200) leaves httpBaseUrl at the convict default http://localhost:3100, so the server listens on one port while getUploadUrl / getSignedUrl / getPublicUrl return URLs on another.

The constructor fallback in local/index.ts (httpBaseUrl ?? \http://localhost:${httpPort}\``) never helps after a normal config load, because convict always supplies the string default.

modules/storage/src/providers/local/index.ts
```diff
@@ -34,11 +78,106 @@ export class LocalStorage implements IStorageProvider {
...
+        const disposition = constructDispositionHeader(baseName, {
+          download,
+          fileName: fileNameParam,
+        });
+        if (disposition) {
+          res.setHeader('Content-Disposition', disposition);
+        }
+        createReadStream(filePath).pipe(res);
+      } else if (req.method === 'PUT') {
+        mkdirSync(dirname(filePath), { recursive: true });
```
ioanniskemerlis: mkdirSync sits outside any try/catch, while writeStream errors are handled below. If directory creation fails (EACCES, ENOSPC, read-only mount), this throws synchronously in the request callback and the client may hang with no HTTP response.

modules/storage/src/Storage.ts
```diff
@@ -221,6 +221,9 @@ export default class Storage extends ManagedModule<Config> {
       this._storageAuthzResourceDispose?.();
       this._storageAuthzResourceDispose = null;
       this.disposeDeclaredPeerWatches();
+      if (this.storageProvider?.dispose) {
+        await this.storageProvider.dispose();
+      }
```
ioanniskemerlis, line 239: Before this PR, local getUploadUrl / getSignedUrl / getPublicUrl threw “not implemented,” so there was no live URL to hand out after storage was turned off.

Now dispose() only closes the embedded server (closeAllConnections + server.close, _httpServer = null). It does not disable URL generation and those methods still return ${_httpBaseUrl}/... and never check whether the server is still running.
Handlers keep the same provider instance, so after active: false a URL RPC can still succeed with e.g. http://localhost:3100/container/file.pdf while nothing is listening.

have getUploadUrl / getSignedUrl / getPublicUrl reject after dispose (or when _httpServer is null), and/or
reject URL RPCs in the handlers when !config.active

Right now deactivate stops the server but still advertises reachable URLs.

modules/storage/src/config/config.ts
ChrisPdgn: Good catch. Fixed by changing the convict default for `httpBaseUrl` to an empty string and using `||` instead of `??` in the constructor so it auto-derives from `httpPort` when unset. `httpBaseUrl` remains configurable for Docker/reverse-proxy scenarios — the doc now explains when to override it vs leave it empty.

modules/storage/src/providers/local/index.ts
ChrisPdgn: Agreed. Wrapped `mkdirSync` in a try/catch in the PUT handler so directory creation failures return a 500 instead of throwing uncaught in the request callback.

modules/storage/src/Storage.ts
ChrisPdgn: Valid point. Added a guard in `getUploadUrl`, `getSignedUrl`, and `getPublicUrl` that rejects when `_httpServer` is null (i.e. after `dispose()`). The framework already stops routing to a NOT_SERVING module, but this protects against races and direct provider calls.

kkopanidis (changes requested):
We need to support some sort of authentication mechanism in the non-public URLs of the provider. For example the "upload" URL part needs something too. We can somehow probably "re-use" authentication's JWT

ChrisPdgn:
Good point on needing auth on the non-public local storage URLs (especially upload). We looked at reusing the authentication module's JWT, but went with **HMAC-SHA256 signed URLs** instead — the same pattern S3/Azure/GCS use for presigned URLs.

**What we implemented:**
- `getUploadUrl()` and `getSignedUrl()` now return URLs with `sig` and `expires` query params
- The embedded HTTP server (port 3100) validates the signature and expiry before serving
- **PUT without a valid signature → 403** (upload URLs are protected)
- **GET with a valid signed URL → serves the file** (private download URLs are protected)
- `getPublicUrl()` stays **unsigned** — public files are meant to be openly accessible, and the URL is stored permanently in the DB

**Why HMAC instead of auth JWT:**
1. **Matches production semantics** — cloud providers don't use user JWTs for presigned URLs; they use HMAC signatures scoped to a specific path + HTTP method + expiry. Local dev should behave the same way.
2. **No cross-module dependency** — the storage provider doesn't need access to authentication's `jwtSecret` or grpcSdk. It's self-contained.
3. **Operation-scoped** — a GET signature can't be used for PUT and vice versa. JWTs don't carry that constraint.
4. **Time-limited by design** — uses the same 1-hour `SIGNED_URL_EXPIRY` constant as AWS/Azure/Aliyun. JWTs have their own longer-lived expiry tied to user sessions.
5. **No JWT in URLs** — putting bearer tokens in query params would leak them in server logs, browser history, and referrer headers. HMAC signatures are opaque and purpose-built for URL-based auth.

The signing secret is ephemeral per provider instance (generated on startup). If the server restarts, outstanding signed URLs expire — same as when cloud signing keys rotate. Upload and download URLs are always generated on-demand through Conduit's authenticated endpoints, so this is fine in practice.

modules/storage/src/providers/local/index.ts
```diff
@@ -34,11 +82,157 @@ export class LocalStorage implements IStorageProvider {
...
+      if (req.method === 'PUT' && !hasSig) {
+        res.writeHead(403);
+        return res.end('Signature required');
+      }
+      if (hasSig && !this.verifyRequest(req.method!, relativePath, sig!, expires!)) {
+        res.writeHead(403);
+        return res.end('Invalid or expired signature');
+      }
+
+      if (req.method === 'GET') {
```
ioanniskemerlis, line 169: The HMAC approach makes sense and I agree with the rationale vs JWT, its good. But the GET handler doesn’t enforce it for private files.

In the comment you made in the PR, it says private downloads are protected by signed URLs, but the server also serves any existing file on unsigned GET (line 138). So getSignedUrl() works, but it’s optional, predictable paths bypass it entirely. 

This is far fetched (very! port should probably be forwarded in normal circumstances, but i dont know what happens in a cafeterias wifi network for example, whare others are in the same network) but it is worth fixing.
The server is still listening on all interfaces, so any path that exposes 3100 (Docker, cloud, misconfig) makes every private file downloadable without a signature. Anything that can reach the machine on port 3100 can hit the server and download private files by path, bypassing Conduit auth and signed URLs entirely e.g. a lab PDF uploaded as isPublic: false is still fetchable at GET /conduit/.../file.pdf with no signature.

Require a valid signature on all GETs except URLs from getPublicUrl() for actually-public files. That matches the design you described, unsigned only for public, signed for private.

modules/storage/src/providers/local/index.ts
```diff
@@ -34,11 +82,157 @@ export class LocalStorage implements IStorageProvider {
...
+          res.writeHead(500);
+          res.end(err.message);
+        });
+      } else {
+        res.writeHead(405);
+        res.end('Method not allowed');
+      }
+    });
+
+    this._httpServer.on('error', err => {
```
ioanniskemerlis, line 269: On listen failure (e.g. EADDRINUSE), this handler only logs,  _httpServer stays non-null, so getUploadUrl / getSignedUrl / getPublicUrl still succeed and return URLs to a server that isn’t accepting connections.

The mkdirSync try/catch you did above,  is a good fix for a different issue (PUT dir creation). This one needs _httpServer = null (or a _listening flag set only in the listen callback) on bind error, so URL generation fails fast when the server didn’t start.

modules/storage/src/providers/local/index.ts
```diff
@@ -172,12 +364,25 @@ export class LocalStorage implements IStorageProvider {
     });
   }
 
-  getPublicUrl(_fileName: string, _containerIsPublic?: boolean): Promise<string | Error> {
-    throw new Error('Method not implemented!');
+  getPublicUrl(fileName: string, _containerIsPublic?: boolean): Promise<string | Error> {
```
ioanniskemerlis, line 456: containerIsPublic is ignored,  we always return a 99-year signed URL. Should we return an unsigned URL for public containers and a long-lived signed URL only for public files in private containers , like azure does? . Works fine as-is, but consider it

ChrisPdgn: Intentionally kept as always-signed — the local HTTP server enforces signatures on all requests (no bucket-level ACLs), matching AWS/GCS behavior. A 99-year signed URL is functionally equivalent to unsigned from the client's perspective.


Christina: "Re bro einai ena feature gia na trexoume local storage poso pio niania na to kanoume pia" ΙΝΑΦΙΣΙΝΑΦ🥲

ioanniskemerlis: fr fr no cap, lgtm, resolved

ioanniskemerlis (approved):

kkopanidis (approved):

PR #1567, opened 2026-07-31
fix(communications): guard template list handlers when params are missing

CluelessBiker:
## Summary
- Defaults `call.request.params` to `{}` in `getTemplates` and `getCommunicationTemplates`.
- Prevents destructuring `undefined` when list requests arrive without query parameters.

## Why
`GET /email/templates` with no query string could throw before the handler ran, returning a 500 instead of an empty or paginated result. This hardens the admin API against a class of request-shape failures seen when the Conduit UI loads templates on page entry.

## Related PR
- Conduit UI: https://github.com/ConduitPlatform/Conduit-UI/pull/301

## Test plan
- [ ] Call `GET /email/templates` with no query params — returns 200 with `templateDocuments` and `count`.
- [ ] Call `GET /communications/templates` with no query params — same behavior.
- [ ] Existing skip/limit/search filtering still works.

### DIFF
--- a/modules/communications/src/admin/index.ts
+++ b/modules/communications/src/admin/index.ts
@@ -211,16 +211,15 @@ export class AdminHandlers {
   async getCommunicationTemplates(
     call: ParsedRouterRequest,
   ): Promise<UnparsedRouterResponse> {
-    const { sort } = call.request.params;
-    const { skip } = call.request.params ?? 0;
-    const { limit } = call.request.params ?? 25;
+    const params = call.request.params ?? {};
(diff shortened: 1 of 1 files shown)


kkopanidis (dismissed):
Removing the reasonable defaults of skip 0 and limit 25 is an unwanted behavior. The APIs provide this default so users cannot accidentally request very large chunks of data. The rest of the parameters were already guarded, not so cleanly of course, but we must preserver the defaults

kkopanidis (approved):

PR #1568, opened 2026-07-31
fix(platform): audit remediation for caching, rate limiting, and ops

CluelessBiker:
## Summary

Follow-up fixes from a full-stack audit of Conduit and Conduit-UI. This PR hardens the backend around caching, rate limiting, error responses, and local Docker setup.

- **Route cache invalidation** — When schemas, documents, or custom endpoints change, related REST and GraphQL route caches are cleared so clients do not see stale data after writes.
- **Rate limiter fail-closed** — If Redis is unavailable, rate limiting now returns 429 instead of silently allowing traffic through. OPTIONS preflight requests are skipped.
- **409 for duplicate resources** — `ALREADY_EXISTS` gRPC errors are mapped to HTTP 409 instead of a generic 500.
- **`GET /live` health endpoint** — Adds a lightweight liveness route for load balancers and ops checks.
- **Docker cleanup** — Removes tracked `docker/.env` (replaced with `.env.example`), fixes the authorization service port in `docker-compose.yml`, and ignores `.pnpm-store/` and build artifacts.

## Why

Stale route caches were the highest-risk finding: admin changes to schemas or endpoints could leave cached responses in place until TTL expiry. Rate-limiter bypass on Redis failure and missing health checks were medium-risk operational gaps. Tracking `docker/.env` in git was a security hygiene issue.

## Related PRs

- **Conduit-UI (admin panel):** [Conduit-UI PR #302](https://github.com/ConduitPlatform/Conduit-UI/pull/302) — audit remediation for auth, errors, and environment handling on the UI side. Land both PRs together for the full audit fix.
- **Conduit (backend):** [Conduit PR #1567](https://github.com/ConduitPlatform/Conduit/pull/1567) — communications template params guard; no file overlap; both should land.

## Test plan

- [ ] Create or update a schema/document and confirm cached `GET /database/...` responses refresh without waiting for TTL
- [ ] Call `GET /live` on the admin API — expect 200
- [ ] Trigger a duplicate create — expect 409, not 500
- [ ] With Redis down, confirm rate-limited routes return 429
- [ ] `docker compose up` with `.env` copied from `.env.example` — authorization service starts on the correct port

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,18 @@
 .turbo
 .DS_Store
 **/env/
+
+# Environment / secrets (never commit)
+docker/.env
+.env
+.env.*
+!.env.example
+
 pnpm-debug.log*
+
+.pnpm-store/
+
+# tsup config build artifacts (source of truth is tsup.config.ts)
+libraries/grpc-sdk/tsup.config.js
+libraries/grpc-sdk/tsup.config.js.map
+libraries/grpc-sdk/tsup.config.d.ts
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -1,10 +1,10 @@
 # Target Container Tags
-IMAGE_TAG="latest"
+IMAGE_TAG="0.17.0-alpha.6"
 UI_IMAGE_TAG="latest"
 
 # gRPC Services
 CORE_GRPC_PORT="55152"
-DB_GRPC_PORT_DB="55160"
+DB_GRPC_PORT="55160"
 ROUTER_GRPC_PORT="55161"
 AUTHN_GRPC_PORT="55162"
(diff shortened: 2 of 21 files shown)


kkopanidis:
This PR touches too many things and introducing inconsistencies. Route cache invalidation needs global handling and not just the DB CRUD routes. Rate limiting assumes (incorrectly) that redis will go down, which if it does, it introduces more problems than the rate limit. The assumption is that redis needs an HA deployment, not mitigation on the server side. 409 is OK, /live endpoint is redundant since /ready covers the case. Docker folder's .env is being used by existing scripts and this would break functionality, while the port changes will probably do the same

PR #1586, opened 2026-08-31
fix(authentication)!: harden JWT, Apple identity tokens, and biometric login

kkopanidis:
## Summary
- Generate `accessTokens.jwtSecret` only when empty (existing `'S3CR3T'` installs keep working); pin HS256 on user and admin JWTs.
- Verify Apple `id_token`s with `ES256` (client callback and token-endpoint payload); stop trusting `header.alg`.
- Require `POST /authentication/biometrics/challenge` and a signature over that nonce (breaking for clients that signed `user._id`).
- Generate TOTP QR codes locally as data URLs instead of sending seeds to QuickChart.

## Test plan
- [ ] New install with empty `jwtSecret` persists a random secret; an existing config with `'S3CR3T'` is left unchanged
- [ ] Access/admin tokens still verify after login; forged `alg` other than HS256 is rejected
- [ ] Apple Sign In (web + native) still creates/logs in users; a token with `alg: HS256` is rejected
- [ ] Biometric: challenge then login succeeds; replaying the old `user._id` signature fails
- [ ] Enabling authenticator 2FA shows a QR image without loading quickchart.io
- [ ] CI: authentication + node-2fa tests pass


ChrisPdgn:
~After 30835938579 fixes, I think this is done, but I want to test it first because this looks critical~
Edit: Branch tested

ChrisPdgn (approved):

PR #1596, opened 2026-09-10
feat(authentication): add OAuth sign-in-only mode

kkopanidis:
Adds a per-request OAuth `mode` on login/register entry points so clients can start a flow as **sign-in only** or **login and register**. `both` remains the default.

This is a client-controlled UX switch (login screen vs register-after-terms). It is not a security boundary: anyone who can call the Client API can pass `mode=both`. A config flag that disables OAuth registration entirely was considered and dropped, because that would block legitimate registration after extra steps.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## API

- Param: `mode`
- Values: `signIn` | `both`
- Default: `both` when omitted

Applied on:

- `GET /init/:provider` and `GET /initNative/:provider` (stored on the OAuth state token)
- Redirect `hook` and `POST /native/:provider` (read from state)
- Direct token routes: `POST /google`, `POST /facebook`

When `mode=signIn` and no matching user exists, the request fails with `REGISTRATION_NOT_ALLOWED`. Existing-user login, account linking, and **invite signup** (`invitationToken`) still work. Anonymous-user conversion is treated as registration and is blocked in sign-in-only mode.

Web redirect hooks send the browser back to `customRedirectUri` / provider `redirect_uri` with `conduitCode=REGISTRATION_NOT_ALLOWED`. Native token routes still return 403 JSON.

## How Has This Been Tested?

- [x] Unit tests for `resolveOAuthMode`, `assertOAuthRegistrationAllowed` (including invitation tokens), and redirect handling

<div><a href="https://cursor.com/agents/bc-c135306e-aa7e-4c0f-add0-08b4a1acaf9f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c135306e-aa7e-4c0f-add0-08b4a1acaf9f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


modules/authentication/src/handlers/oauth2/OAuth2.ts
```diff
@@ -315,6 +327,7 @@ export abstract class OAuth2<
       if (!user.isVerified) user.isVerified = true;
       user = await User.getInstance().findByIdAndUpdate(user._id, user);
     } else {
+      assertOAuthRegistrationAllowed(mode);
```
ChrisPdgn: **`invitationToken` + `mode=signIn` rejects a real invite signup.**

This assert runs in the “no User found” branch *before* invite validation (`inviteValidation` / `addUserToTeam` below). A new invitee who starts OAuth with both `invitationToken` and `mode=signIn` gets `REGISTRATION_NOT_ALLOWED` after they already consented at the provider. The invite token is not consumed.

That will happen if an invite landing page reuses the login-screen OAuth helper (same “Continue with Google”, plus the token). Local `POST /local/new` still accepts invites; only OAuth invite signup breaks.

An invite **is** registration. Either:
- treat a present `invitationToken` as registration intent and skip this assert, or
- fail at `/init` when both are sent, so the user never completes Google/Apple consent first.

modules/authentication/src/handlers/oauth2/OAuth2.ts
```diff
@@ -223,6 +232,7 @@ export abstract class OAuth2<
       payload,
       stateToken.data.invitationToken,
       stateToken.data.anonymousUserId,
+      resolveOAuthMode(stateToken.data.mode),
```
ChrisPdgn: **Web redirect `signIn` failures never return to the app.**

On success, `TokenProvider.provideUserTokens` redirects to `customRedirectUri` / `finalRedirect` with tokens. If `createOrUpdateUser` throws `REGISTRATION_NOT_ALLOWED` here, the router returns 403 JSON on `/hook/authentication/{provider}` and `redirectUri` is never used.

That is pre-existing for rare hook errors (inactive user, bad invite), but `mode=signIn` on a login screen makes “this Google account has no User yet” a *normal* outcome of tapping Continue with Google.

Native `POST /google` / `POST /facebook` are fine (the app can handle 403). Redirect `/init/:provider` is not — the browser is stuck on Conduit.

Catch `REGISTRATION_NOT_ALLOWED` in `authorize` (and Apple’s override) and redirect to `customRedirectUri` with `conduitCode=REGISTRATION_NOT_ALLOWED` instead of throwing on the hook. Same issue exists on Apple’s `authorize`.

ChrisPdgn (commented):
Two product holes in `mode=signIn`. Default `both` is unchanged and non-breaking.

ChrisPdgn (approved):
Looks good — both review comments are addressed (`invitationToken` is treated as registration intent; web hooks redirect with `conduitCode=REGISTRATION_NOT_ALLOWED`).

Holding merge until the same fix lands on #1597 so 0.16.x matches this contract.

